PR #131, opened 2021-01-10
Role filter on role governance

jrhender:
Added filter for role on Role Governance tab. Essentially a copy-paste of the other filters.
Adjusted filter column size to fit role filter


kim-energyweb (approved):
LGTM

PR #149, opened 2021-01-29
fix(dashboard-init): restore re-login functionality

jrhender:
Issue was that useMetamaskExtension param was removed so iam-client-lib was throwing exception when receiving receiving WalletProvider type that was not allowed.

Functionality may still not be ideal. Relogin on reload with metamask works. However, walletconnect/KM re-login likely does not. I don't believe it currently works on https://volta-switchboard.energyweb.org (master branch) either but I'm not certain. I'm not sure if the walletConnect library is checking for existing session information from local storage or if we would need to pass in an existing walletConnect session info into iam-client-lib and then re-establish the connection to the session based on that. Needs a bit more investigation on my part...


kim-energyweb (approved):
LGTM

PR #164, opened 2021-02-02
Myen 605 communicating KeyManager new tab

jrhender:
This PR:
- Adds "WalletProvider" to request-claim component (was missing previously)
- Adds additional text to "signature requested" message when doing _initial_ login EW KeyManager. When signing in after a refresh, can't currently display different text as we do not know if the WalletConnect session is with KM or another wallet.

image


kim-energyweb (commented):
I am confirming the messages with Mani & Ian.
Once it is approved, I will merge this PR.

As checked, the codes looks good to me.

jrhender:
Updated text: 
image

manihagh (approved):

PR #173, opened 2021-02-08
MYEN635 Increase timeout for KM signin to 2 minutes

jrhender:
Only KM has 2 minute timeout. Other providers still have 1 minute


src/app/shared/services/iam.service.ts
```diff
@@ -256,7 +257,7 @@ export class IamService {
     }
   }
 
-  private async _displayTimeout(timeout: number, isConnectAndSign?: boolean) {
```
jrhender, line 259: `timeout` removed here as not used by `_displayTimeout`

kim-energyweb (approved):
LGTM

PR #178, opened 2021-02-10
feature/integrate with gnosis

JGiter:
- [x] Inject `SafeAppSdk` if one provided by the page
- [x] When gnosis sdk is provied list of domains is appended with domains owned by safe wallet
- [x] Operations under safe wallet's domains are performed in one step


kim-energyweb (commented):
LGTM

PR #184, opened 2021-02-18
bug/nodejs packages not resolved

JGiter:
Updated iam-client-lib to version which contains fix for not resolved by Webpack nodejs packages and removed not needed anymore browser section in package.json


kim-energyweb (approved):
LGTM

PR #222, opened 2021-04-12
ui fixes

marcin-l-tsh:
* fixed organization searchbar color
* 'Add asset type' button disabled
* fixed bug which blocked user from typing organisation namespace while using uppercase letters


kim-energyweb (changes requested):
`fixed bug which blocked user from typing organisation namespace while using uppercase letters`

Namespaces are always lowercase letters.

kim-energyweb (approved):
LGTM

PR #244, opened 2021-04-29
[SWTCH-854] - Fix Tooltip UI Issue in Enrolment Requests screen

davycastillo:


kim-energyweb (changes requested):
Please update for below files as well:
- asset-list
- asset-ownership-history

PR #256, opened 2021-05-11
update: iam-client-lib dependency.

DawidGil-jlabs:


jrhender:
@DawidGil-jlabs Is this the branch where you were seeing the `403 forbidden` response code? If so, you aren't seeing it in develop branch, right?

DawidGil-jlabs:
@jrhender Yes, it is. Branch with 403 forbidden.

jrhender (approved):

PR #260, opened 2021-05-14
Fix Notification duplicate

VOSpdUkraine:


jrhender:
@VOSpdUkraine Could explain a bit why this change fixes the issue? Thanks 🙏

src/app/layout/header/header.component.ts
```diff
@@ -140,7 +139,9 @@ export class HeaderComponent implements OnInit, OnDestroy {
 
         // Stay in current screen and display user name if available
         this.iamService.userProfile
-            .pipe(takeUntil(this._subscription$))
+            .pipe(
+                skip(1),
```
kim-energyweb, line 143: I suggest to use a flag inside `_initNotificationListeners()` to check if notif listeners are already set. We can't guarantee that it would always be twice. It could be more. So, skipping once is not a good choice. Please let me know on your opinion on this.

kim-energyweb (changes requested):

kim-energyweb (approved):
LGTM

PR #264, opened 2021-05-17
Feat/enable unit tests to run them on pipeline

DawidGil-jlabs:
This PR fixes some of the tests, but most of them are disabled. It is needed so we could start writing unit tests to existing code and add new one while developing on new features.


kim-energyweb (approved):
LGTM

PR #266, opened 2021-05-17
fix(SWTCH-942): assets for did without profile claim do not appear on the list.

DawidGil-jlabs:
Fixes 
1. problem when trying to register new asset and user didn't have user profile claim.
2. now should stop refreshing asset list when closing edit asset dialog component while not updating data.
3. now should stop closing edit asset dialog when clicking on backdrop.


kim-energyweb (approved):
LGTM

PR #269, opened 2021-05-19
SWTCH-890 - Update angular

DawidGil-jlabs:
This PR updates Angular from version 8.0.0 to 10.2.5. This also updates connected packages with angular that are used in project.


Olesia-spd:
SWTCH-890

e2e/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "outDir": "../out-tsc/e2e",
     "module": "commonjs",
-    "target": "es5",
+    "target": "es2018",
```
jrhender, line 6: Just curious, why the upgrade to `es2018` here?

package.json
```diff
@@ -11,8 +11,7 @@
     "start-dev": "ng build --watch",
     "build-prod": "node --max_old_space_size=8048 node_modules/@angular/cli/bin/ng build --prod --aot -c=prod",
     "build": "node --max_old_space_size=8048 node_modules/@angular/cli/bin/ng build --prod --aot -c=dev",
-    "postinstall": "node patch.js && npm run downlevel_gnosis",
-    "downlevel_gnosis": "node ./scripts/downlevel_gnosis",
```
jrhender, line 15: 🥳

jrhender (approved):
Looks good to me

e2e/tsconfig.json
DawidGil-jlabs: It's automated by `ng update`. I didn't done that manually. I think i can rollback this one change, because fresh angular application have es2015

jrhender: Ok, `es2015` sounds good to me if that works

DawidGil-jlabs: oh i see, this is tsconfig for e2e tests. So it's not relevant imo, because we don't use here e2e tests yet. Application tsconfig is not changed and have `"target": "es2015",`

jrhender: Ah fair enough. No worries then 👍

PR #271, opened 2021-05-20
SWTCH-891 change labels owner/offer to depending on event type.

DawidGil-jlabs:
If this is ASSET_TRANSFERRED we should display offered from/owner labels.


jrhender (changes requested):
I don't think this is quite correct. I think for the `ASSET_TRANSFERRED` event, the owner is still the `element.emittedBy`. The screenshot below is taken from running this branch.
The original owner was `0x829b91Fa3e91EA4448365ADA58C7Bad1Ff142866` and the new owner is `0x397a2b6ee5d0D18EFA59C54E2254722405D6bC83`. So I think the correct result should be
```
Offered from 0x829b91Fa3e91EA4448365ADA58C7Bad1Ff142866
Owner 0x397a2b6ee5d0D18EFA59C54E2254722405D6bC83
```
image

DawidGil-jlabs:
@jrhender changes made. Should be correct now. I additionaly split some code into more components + added some unit tests.

jrhender (changes requested):
@DawidGil-jlabs When running the branch and viewing history I know don't see DIDs associated with the events. Can you please check?

image

jrhender (approved):
Looks good, I like the tests 👍

PR #272, opened 2021-05-21
[SWTCH-946] Fix  notification counter

kim-energyweb:
This PR is created by the request of Vadym. He can't login to his GitHub account.


jrhender (approved):
Looks good to me. @DawidGil-jlabs If you've got time to take a look as well, that would be great 🙇

src/app/layout/header/header.component.html
```diff
@@ -79,8 +79,25 @@ <h4 class="mb-0 f-16 color-white">Tasks</h4>
                                             <mat-icon svgIcon="new-claim-icon"></mat-icon>
                                         </div>
                                         <div
-                                            class="media-body d-flex flex-row justify-content-between align-items-center">
```
DawidGil-jlabs, line 82: Why so big spaces? Maybe we should stick with one way, I mean `<div class="">...</div>` without additional enters/tabs/spaces. What do you think?

src/app/layout/header/header.component.ts
```diff
@@ -301,13 +298,24 @@ export class HeaderComponent implements OnInit, OnDestroy {
...
+            this._initZeroCheckerForPendingSyncCount();
+        } catch (e) {
             throw e;
         }
     }
 
-    private async _initAssetsOfferedToMeSyncCount(){
+    private _initZeroCheckerForPendingSyncCount(): void {
+        this.notifService.pendingSyncCount$
+            .pipe(takeUntil(this._subscription$))
```
DawidGil-jlabs, line 309: You can add here filter operator, i mean `filter(count => count === 0)`. That way subscribe doesn't need if statement

DawidGil-jlabs (approved):

src/app/layout/header/header.component.html
```diff
@@ -79,8 +79,25 @@ <h4 class="mb-0 f-16 color-white">Tasks</h4>
...
+                                                class="media-body d-flex flex-row justify-content-between align-items-center">
```
DawidGil-jlabs, line 82: Why so big spaces? Maybe we should stick with one way, I mean <div class="">...</div> without additional enters/tabs/spaces. What do you think?

DawidGil-jlabs:
Yeah, for me it also looks good, that's why i approved it. Just added comments to start discussion :) i think we need to enable tslint and in feature eslint (tslint is deprecated in favor of eslint).

DawidGil-jlabs (approved):

DawidGil-jlabs (approved):

PR #278, opened 2021-05-27
[SWTCH-933] Fix User enrolment should have an empty asset field

VOSpdUkraine:


DawidGil-jlabs (approved):

jrhender (changes requested):
If I go to my "Enrolment Requests", I still see asset DID for ones where the `subject == requester`. I think the logic that you put is the opposite of what it should be. What we want is that don't show the asset DID if subject and requester are the same.

image

jrhender (changes requested):
When I try now, the first row doesn't show the asset DID but the other rows show the asset DID even though the asset and owner DIDs are the same. I think therefore that it is still not correct. @VOSpdUkraine could you confirm that it is working for you?
image

jrhender:
@VOSpdUkraine Just a thought, but it seems to me like this might be a good feature as well to have a frontend unit test. Perhaps @DawidGil-jlabs can help you with this if necessary.

DawidGil-jlabs:
Yeah, i can help with testing. Unfortunately some of our components are too big and have to many dependencies to test them easly. 
However we can always create a function and test that function. Then import it into that component and use it inside a method. Pure functions are very easy to test. 
Or if you don't want to create a function, you can always create a pipe which will be comparing two strings. Depending on that pipe will display text or empty string.

VOSpdUkraine:
@DawidGil-jlabs 
we need not only to hide the text but also the icon, title and text
does it make sense in two places to write a structured directive and the logic here is very simple

@jrhender 
yes you are right, I provided the wrong condition: if "subject !== requester" is hide and if "subject === requester" is show

I fixed

jrhender (approved):

DawidGil-jlabs (approved):
I approved, but what do you think about moving logic from html template to ts/js files?
Good method names could save some time and you have better understanding what's happening in html.

PR #279, opened 2021-05-27
feat(SWTCH-895): implement addition of verification methods to asset did document.

DawidGil-jlabs:


jrhender (changes requested):
@davycastillo @DawidGil-jlabs Sorry, maybe it wasn't clear from the story description. But the user definitely needs to be able to provide the actual public key string as well as the key type. Currently looks like can only provide the type:
image

DawidGil-jlabs:
Updated view with new input field for public key with validation. Added validation for Select component that it is required.

Displayed public key is truncated, so while mouse hovering there is showed tooltip with full string. We can also copy to clipboard public key.
verification
I hope right now it better suits requirements.

jrhender (approved):
This is looking good 👍  I tested it out, seems to work well. We could potentially improve the key value validation but this looks good for now.

DawidGil-jlabs:
Yeah, i was thinking about additional validation, but i didn't know what string we should accept so i implemented only default ones.

PR #280, opened 2021-05-27
fix(SWTCH-930): hide the Enrolment button on offered and previously owned pages only when asset don't have enrolments.

DawidGil-jlabs:


jrhender (approved):

jrhender:
@DawidGil-jlabs I think this PR is probably fine, but IMO we should wait to merge until it actually works fully. I.e. this scenario should work:

1. Create a new asset and request a role for this asset
2. Offer the asset to another account
3. View asset on “offered to me” tab of other account → enrolments button should show
4. Click on enrolments button. → No enrolments are shown ❌  but enrolment request should be shown

VOSpdUkraine (approved):

jrhender (approved):
Unable to fully test as I think that enrolments are broken but seems to work as far as I can tell

PR #286, opened 2021-05-28
[SWTCH-961] Issue on-chain credential

VOSpdUkraine:
Cannot be checked before it is done "BE"

### DIFF
--- a/src/app/routes/enrolment/view-requests/view-requests.component.html
+++ b/src/app/routes/enrolment/view-requests/view-requests.component.html
@@ -62,7 +62,7 @@
             <mat-icon>info</mat-icon><span class="f-14">Role did not require any field for enrolment request.</span>
           </div>
         </div>
-        <div class="col-lg-12 mt-2">
+        <div class="col-lg-12 mt-2" *ngIf="claim && claim.registrationType === 'OnChain'">
           <div class="card card-info d-flex flex-row justify-content-start align-items-center shadow-none d-flex mb-3">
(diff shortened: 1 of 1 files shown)


DawidGil-jlabs (approved):
This should only display information when offchain and onchain is enabled, right? If it's only that, then it looks fine for me.

jrhender:
IMO, would be good to integrate https://github.com/energywebfoundation/iam-client-lib/pull/189 and test issuance before merging this PR

jrhender (changes requested):
@VOSpdUkraine Could you integrate `iam-client-lib` that has the on-chain role issuance (>3.0.0-alpha.21) and rebase/merge with the latest `development` branch please so that we can test the request+issuance end-to-end in this PR?

jrhender:
@VOSpdUkraine When testing, I'm noticing that even if both "On-chain" and "Off-chain" registrationTypes are listed in the request, the message about two signatures isn't shown. For example, in the screenshot below you can see that both `Off-chain` and `On-chain` are coming back from the server but the message still isn't displayed. Could double check why this might be?
To test locally end-to-end, you would need to run a local cache-server using branch in this PR https://github.com/energywebfoundation/iam-cache-server/pull/138 

image

jrhender (approved):
@VOSpdUkraine Thanks for the change! Getting closer. I was able to issue an on-chain role! Still, following acceptance criteria is not met:
> Credential is being issued (i.e. should not show when asset owner is viewing request as is not a relevant message to asset owner)
For instance, I see the multi-signature message when viewing my enrolment requests. It should only be shown at the time of issuance.

Still, I think we should merge this change so that other folks can test in dev environment and we can fix in another PR. How does that sound?

PR #287, opened 2021-05-28
SWTCH-968 Integrate new org/app/role definition creation into SB

jrhender:
Minor changes need to create and update Roles, Apps, Orgs using the new RoleDefinitionResolver smart contract:

- feat(new-role.component): use number version. As number was previously strings of the format "1.0.0", need to convert into a number as we are now using integers for role versions

Still in DRAFT as need to integrate updated iam-client-lib first

### DIFF
--- a/src/app/routes/applications/new-role/new-role.component.ts
+++ b/src/app/routes/applications/new-role/new-role.component.ts
@@ -209,10 +209,6 @@ export class NewRoleComponent implements OnInit, AfterViewInit, OnDestroy {
       let arrParentNamespace = this.origData.namespace.split(ENSNamespaceTypes.Roles);
       let parentNamespace = arrParentNamespace[1].substring(1);
 
-      // Construct Version
-      const arrVersion = typeof(def.version) === 'string' ? parseInt(def.version.split('.')[0], 10) : def.version;
-      const version = arrVersion + 1;
-
       // Construct Fields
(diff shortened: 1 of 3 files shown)


src/app/routes/applications/new-role/new-role.component.ts
```diff
@@ -210,8 +210,8 @@ export class NewRoleComponent implements OnInit, AfterViewInit, OnDestroy {
       let parentNamespace = arrParentNamespace[1].substring(1);
 
       // Construct Version
-      let arrVersion = def.version.split('.');
-      let version = `${parseInt(arrVersion[0]) + 1}.0.0`;
+      const arrVersion = typeof(def.version) === 'string' ? parseInt(def.version.split('.')[0], 10) : def.version;
```
DawidGil-jlabs, line 213: Could you extract this to a method? So those two lines disappear from here and in patching form you can assign result of this method to `version`.

jrhender: @DawidGil-jlabs thanks, good tip. I updated the PR. Can you take a look and see if it is what you had in mind when you have time?

DawidGil-jlabs: Yes, this is what i had in mind. Thanks!

DawidGil-jlabs (approved):
Lgtm

DawidGil-jlabs (approved):

PR #288, opened 2021-05-31
[SWTCH-859] fixes to verification methods

DawidGil-jlabs:
Fixes to adding verification methods:
1. Added validation to accept Hex inputs as public key. Thanks to @JGiter for providing regex.
2. Change labels
3. Change toastr message to display Public key instead of Did.
4. Removed not needed css classes.

Additionaly added directive for copying to clipboard and displaying message with toastr - this can be used in multiple places where clipboard functionality is needed.

### DIFF
--- a/src/app/routes/assets/verification-method/verification-method.component.ts
+++ b/src/app/routes/assets/verification-method/verification-method.component.ts
@@ -77,7 +77,7 @@ export class VerificationMethodComponent implements OnInit {
   }
 
   private copied(): void {
-    this.toastr.success('Did successfully copied to clipboard.');
+    this.toastr.success('Public key successfully copied to clipboard.');
   }
 
 }
--- a/src/app/routes/assets/verification-method/verification-method.component.ts
+++ b/src/app/routes/assets/verification-method/verification-method.component.ts
@@ -5,7 +5,6 @@ import { FormControl, Validators } from '@angular/forms';
 import { PageEvent } from '@angular/material/paginator';
 import { AlgorithmsEnum } from '../models/algorithms.enum';
 import { ToastrService } from 'ngx-toastr';
-import { textDefaultsValidators } from '../../../utils/validators/text-defaults-validators';
 
 export interface PublicKey {
   publicKeyHex: string;
@@ -24,7 +23,7 @@ export class VerificationMethodComponent implements OnInit {
   dataSource: PublicKey[] = [];
   selectControl = new FormControl('', [Validators.required]);
   publicKey = new FormControl('',
-    textDefaultsValidators
+    [Validators.required, Validators.pattern(new RegExp('0x[A-Fa-f0-9]{66}'))]
   );
   selectOptions = Object.entries(AlgorithmsEnum);
   private publicKeys;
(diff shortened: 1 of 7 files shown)


src/app/utils/validators/is-hex.validator.spec.ts
```diff
@@ -0,0 +1,42 @@
+import { isHexValidator } from './is-hex.validator';
+import { FormControl } from '@angular/forms';
+
+describe('tests for isHexValidator', () => {
+  const hexWithoutBeggining66 = 'aaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeeeffffffffff123456';
+  const hexWithoutBeggining130 = 'aaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeeeffffffffff123456aaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeeeffffffffff1234';
```
jrhender, line 6: Typo in "Beginning"

jrhender (approved):

src/app/utils/validators/is-hex.validator.spec.ts
DawidGil-jlabs: haha, thanks :D updated :)

src/app/routes/assets/verification-method/verification-method.component.ts
```diff
@@ -61,6 +59,19 @@ export class VerificationMethodComponent implements OnInit {
     });
   }
 
+  getPublicKeyErrorMsg() {
+    debugger;
```
VOSpdUkraine, line 63: ?

src/app/utils/validators/is-hex.validator.ts
```diff
@@ -0,0 +1,17 @@
+import { AbstractControl } from '@angular/forms';
+
+export function isHexValidator(control: AbstractControl) {
+  if (!control.value) {
+    return null;
+  }
+  const hexRegex = new RegExp('^(0x)([0-9a-fA-F])');
```
VOSpdUkraine, line 7: it will be better if we create a RegEx when the file is initialized and not every time the method is called

src/app/utils/validators/is-hex.validator.ts
```diff
@@ -0,0 +1,17 @@
...
+  // 0x + 66 = 68, 0x + 130 = 132.
+  const validLength = control.value.length === 68 || control.value.length === 132;
```
VOSpdUkraine, line 9: 68 and 132 maybe will be better move into global const ?

src/app/routes/assets/verification-method/verification-method.component.ts
DawidGil-jlabs: Yeah, forgot to remove that. Thanks for finding it out.

src/app/utils/validators/is-hex.validator.ts
DawidGil-jlabs: updated. Please check if this is what you expected.

VOSpdUkraine (approved):

PR #291, opened 2021-06-02
[SWTH-962] update implement UI changes for Set Field step

VOSpdUkraine:


DawidGil-jlabs (commented):
We can't specify min/max values for boolean type. I think there is no point to display table in that case.
Zrzut ekranu 2021-06-8 o 11 45 13

DawidGil-jlabs:
After quick talk with @davycastillo he suggested: `Also min value and max value should be hidden when not specified or empty`

DawidGil-jlabs (approved):
Tested. Looks good to me.

PR #294, opened 2021-06-07
[SWTCH-989] Add disable menu item

davycastillo:
Add disable style to action menu item.

### DIFF
--- a/src/app/routes/applications/governance-list/governance-list.component.scss
+++ b/src/app/routes/applications/governance-list/governance-list.component.scss
@@ -1,0 +1,4 @@
+.mat-menu-item[disabled], .mat-menu-item[disabled]::after,
+.mat-menu-item[disabled] .mat-icon-no-color, .mat-menu-item-submenu-trigger[disabled]::after  {
+    color: rgba(255,255,255,0.3) !important;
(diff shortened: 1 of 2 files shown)


src/app/routes/applications/governance-list/governance-list.component.scss
```diff
@@ -0,0 +1,4 @@
+.mat-menu-item[disabled], .mat-menu-item[disabled]::after,
+.mat-menu-item[disabled] .mat-icon-no-color, .mat-menu-item-submenu-trigger[disabled]::after  {
```
DawidGil-jlabs, line 2: Is it needed to add: `.mat-menu-item-submenu-trigger[disabled]::after` and `.mat-menu-item[disabled]::after` ? I don't see a case for it that's why i'm asking :)

DawidGil-jlabs (approved):

src/app/routes/applications/governance-list/governance-list.component.scss
davycastillo: Yeah, somehow it needs to tag these classes to custom change the color of the text to disabled

PR #296, opened 2021-06-09
SWTCH-940 implement state management

DawidGil-jlabs:
State management implementation for updating user claims/profile with tests for that part.


src/app/layout/header/header.component.html
```diff
@@ -199,15 +197,17 @@ <h4 class="mb-0 f-16 color-white">Notification</h4>
...
-                height="44" alt="avatar">
-            <div class="label-role" *ngIf="!userName">Manage Profile</div>
-            <div class="label-role" *ngIf="userName"> {{ userName }} </div>
-        </div>
-        <div class="d-flex align-items-center opacity-50 text-truncate-middle did-wrapper mx-3 mb-3"
-            matTooltip="Copy to clipboard" [cdkCopyToClipboard]="userDid" (cdkCopyToClipboardCopied)="didCopied()">
-            <div class="f-14">{{userDid | didFormatMinifier}}</div>
-            <mat-icon class="ml-2">content_copy</mat-icon>
+                 height="44" alt="avatar">
+            <div class="label-role">{{ userName$ | async}}</div>
```
VOSpdUkraine, line 201: duplicate 'async' (185/201), it will be better to combine 'async'

src/app/layout/header/dialog-user/dialog-user.component.ts
```diff
@@ -1,102 +1,72 @@
...
-            this.maxDate = today;    
-    }
-
-    async ngOnInit() {
-        this.loadingService.show();
+  constructor(
+    public dialogRef: MatDialogRef<DialogUserComponent>,
+    private fb: FormBuilder,
+    private store: Store<UserClaimState>) {
+    this.profileForm = fb.group({
```
VOSpdUkraine, line 26: we need to avoid complex actions in the constructor

I think, it will be better:
export class DialogUserComponent implements OnInit, OnDestroy {   
.....
 public profileForm = this.fb.group({
......

src/app/layout/header/dialog-user/dialog-user.component.ts
```diff
@@ -1,102 +1,72 @@
...
+        Validators.maxLength(500),
+        Validators.required
+      ])]
+    });
 
-        // Get User Claims
-        let data: any[] = await this.iamService.iam.getUserClaims();
-        // console.log('getUserClaims()', JSON.parse(JSON.stringify(data)));
+    const today = new Date();
+    today.setFullYear(today.getFullYear() - 18);
```
VOSpdUkraine, line 39: magic number "18"

src/app/layout/header/dialog-user/dialog-user.component.ts
DawidGil-jlabs: Yeah, i didn't wrote that. And i don't want to refactor everything at once. However this is small change so i will do that now.

src/app/layout/header/dialog-user/dialog-user.component.ts
DawidGil-jlabs: Yeah, not my code again. Add some info to this...

src/app/layout/header/header.component.html
DawidGil-jlabs: You want to make one async like this: `<ng-container *ngIf="userName$ | async as userName"> everything from component </ng-container>` ?

jrhender (approved):
Looks good to me. Tested as well, no issues found 👍

PR #299, opened 2021-06-10
fix(SWTCH-990): not working back navigation when navigating from a url with redirect inside.

DawidGil-jlabs:
Previous solution wasn't working when navigating to asset enrolments from enrol/dashboard redirect. Back button was navigating to that page, but it was immediately redirecting to asset enrolments again. In that case back button will navigate to `asset` url which is my asset list.

### DIFF
--- a/src/app/routes/assets/asset-enrolment-list/asset-enrolment-list.component.ts
+++ b/src/app/routes/assets/asset-enrolment-list/asset-enrolment-list.component.ts
@@ -4,7 +4,7 @@ import { ActivatedRoute } from '@angular/router';
 import { Subject } from 'rxjs';
 import { takeUntil } from 'rxjs/operators';
 import { EnrolmentListComponent } from '../../enrolment/enrolment-list/enrolment-list.component';
-import { Location } from '@angular/common';
+import { UrlService } from '../../../shared/services/url-service/url.service';
 
 @Component({
   selector: 'app-asset-enrolment-list',
@@ -28,7 +28,7 @@ export class AssetEnrolmentListComponent implements OnInit, OnDestroy {
   private subscription$ = new Subject();
 
   constructor(private activatedRoute: ActivatedRoute,
-              private location: Location) {
+              private urlService: UrlService) {
   }
 
   ngOnDestroy(): void {
@@ -51,6 +51,18 @@ export class AssetEnrolmentListComponent implements OnInit, OnDestroy {
   }
 
   back() {
-    this.location.back();
+    this.urlService.previous.pipe(
+      takeUntil(this.subscription$)
+    ).subscribe(url => {
+      this.navigateBack(url);
+    });
+  }
+
+  navigateBack(url: string): void {
+    if (url.includes('returnUrl')) {
+      this.urlService.goTo('assets');
+      return;
+    }
+    this.urlService.back();
   }
 }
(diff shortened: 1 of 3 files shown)


src/app/shared/services/url-service/url.service.ts
```diff
@@ -0,0 +1,44 @@
...
+import { filter } from 'rxjs/operators';
+import { BehaviorSubject, Observable } from 'rxjs';
+import { Location } from '@angular/common';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class UrlService {
+  private previousUrl = new BehaviorSubject<string>(null);
+  private currentUrl = new BehaviorSubject(null);
```
jrhender, line 12: Why is `previousUrl` constrained to `string` but `currentUrl` isn't?

src/app/routes/assets/asset-enrolment-list/asset-enrolment-list.component.ts
```diff
@@ -51,6 +51,18 @@ export class AssetEnrolmentListComponent implements OnInit, OnDestroy {
...
-    this.location.back();
+    this.urlService.previous.pipe(
+      takeUntil(this.subscription$)
+    ).subscribe(url => {
+      this.navigateBack(url);
+    });
+  }
+
+  navigateBack(url: string): void {
+    if (url.includes('returnUrl')) {
```
jrhender, line 62: I think a comment could be valuable here, as I think that it is not immediately obvious why this logic is in place
```suggestion
    // 'returnUrl' is taken as an indicator back() would trigger an loop back to asset-enrolment
    if (url.includes('returnUrl')) {
```

src/app/routes/assets/asset-enrolment-list/asset-enrolment-list.component.ts
```diff
@@ -51,6 +51,18 @@ export class AssetEnrolmentListComponent implements OnInit, OnDestroy {
...
   back() {
-    this.location.back();
+    this.urlService.previous.pipe(
+      takeUntil(this.subscription$)
+    ).subscribe(url => {
+      this.navigateBack(url);
+    });
+  }
+
+  navigateBack(url: string): void {
```
jrhender, line 61: Should `navigateBack()` be private? If not, could the naming be improved? The names don't imply a difference in the two functions, to me at least.

jrhender (approved):
Just a few comments but looks good overall

src/app/shared/services/url-service/url.service.ts
```diff
@@ -0,0 +1,44 @@
...
+  get previous(): Observable<string> {
+    return this.previousUrl.asObservable();
+  }
+
+  get current() {
+    return this.currentUrl.asObservable();
+  }
+
+  constructor(private router: Router,
+              private location: Location) {
```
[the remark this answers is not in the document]
DawidGil-jlabs: I don't see a case when someone would overwrite those properties. I saw and i wrote sometimes `readonly` with class parameters, but not when declaring properties in constructor.

src/app/routes/assets/asset-enrolment-list/asset-enrolment-list.component.ts
DawidGil-jlabs: Yes, it should be private. Maybe also rename it into `navigateBackHandler` ?

src/app/shared/services/url-service/url.service.ts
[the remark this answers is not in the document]
jrhender: Fair enough, very unlikely, I agree

src/app/routes/assets/asset-enrolment-list/asset-enrolment-list.component.ts
jrhender: Sounds good

PR #302, opened 2021-06-16
[swtch 971] improve notification logic in the system

VOSpdUkraine:


src/app/layout/header/header.component.ts
```diff
@@ -56,6 +56,16 @@ export class HeaderComponent implements OnInit, OnDestroy {
     isLoadingNotif = true;
     userName$ = this.store.select(userSelectors.getUserName).pipe(map(value => value ? value : 'Manage Profile'));
     userDid$ = this.store.select(userSelectors.getDid);
+    notificationNewItems = 0;
+    notificationList$: Observable<SwitchboardToaster[]> = this.toastr.getMessageList()
+        .pipe(tap(items => {
+            this.notificationNewItems = 0;
```
DawidGil-jlabs, line 62:
```suggestion
            this.notificationNewItems = items.filter(item => item.isNew).length;
```
What do you think?

src/app/routes/assets/verification-method/verification.service.spec.ts
```diff
@@ -16,7 +16,7 @@ describe('VerificationService', () => {
       providers: [
         { provide: IamService, useValue: iamServiceSpy },
         { provide: LoadingService, useValue: loadingServiceSpy },
-        { provide: ToastrService, useValue: toastrSpy }
+        { provide: SwitchboardToasterService, useValue: toastrSpy }
```
DawidGil-jlabs, line 19: I think you should also adjust toastrSpy, because it is right now `const toastrSpy = jasmine.createSpyObj('ToastrService', ['success']);`

src/app/shared/services/switchboard-toaster.service.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { SwitchboardToasterService } from './switchboard-toaster.service';
+import { ToastrModule, ToastrService } from 'ngx-toastr';
+
+describe('SwitchboardToasterService', () => {
+  let service: SwitchboardToasterService;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      providers: [ToastrService],
```
DawidGil-jlabs, line 11: I think you should provide a Toastr mock, not real service. I think we should test our services, not libraries.

src/app/shared/services/switchboard-toaster.service.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+      imports: [ToastrModule.forRoot()]
+    });
+    service = TestBed.inject(SwitchboardToasterService);
+  });
+
+  it('should be created', () => {
```
DawidGil-jlabs, line 17: It would be great if you add some tests to this service, so in future nobody break this by refactor/adding new functionalities.

src/app/shared/services/switchboard-toaster.service.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export interface SwitchboardToaster {
+    message: string;
+    type: string;
+    isNew: boolean;
+}
+@Injectable({
+    providedIn: 'root'
+})
+export class SwitchboardToasterService {
```
DawidGil-jlabs, line 21: I'm not sure if this is a typo or on purpose, because this is wrapper for Toastr, not Toaster :) And for clarity I have nothing against this name! I'm just curious :D

src/app/shared/services/switchboard-toaster.service.ts
```diff
@@ -0,0 +1,76 @@
...
+        this.updateMessageList(message, MessageType.warning);
+    }
+
+    private updateMessageList(message: string, type: string): void {
+        let list = this.massageList.getValue();
+        list.unshift({message, type, isNew: true});
+        if (list.length > 20) {
+            list = list.slice(0, 20);
+        }
+        this.massageList.next(list);
```
DawidGil-jlabs, line 74: Maybe
```suggestion
        this.massageList.next(list.slice(0,20));
```
That way we can remove if statement. This should work always, because when we get smaller list than 20 elements then slice will return only those elements.

src/app/shared/services/switchboard-toaster.service.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    warning(message?: string, title?: string, override?: Partial<IndividualConfig>): any {
+        this.toaster.warning(message, title, override);
+        this.updateMessageList(message, MessageType.warning);
+    }
+
+    private updateMessageList(message: string, type: string): void {
+        let list = this.massageList.getValue();
+        list.unshift({message, type, isNew: true});
+        if (list.length > 20) {
```
DawidGil-jlabs, line 71: > Magic number 20

:D

src/app/shared/services/switchboard-toaster.service.ts
```diff
@@ -0,0 +1,76 @@
...
+    getMessageList(): Observable<SwitchboardToaster[]> {
+        return this.massageList.asObservable();
+    }
+
+    reset(): void {
+        this.massageList.next([]);
+    }
+
+    readAllItems(): void {
+        this.massageList.next(this.massageList.getValue().map(item => {
```
DawidGil-jlabs, line 36: What do you think about not mutating data? So it would be
```suggestion
        this.massageList.next(this.massageList.getValue().map(item => ({...item, isNew: false}))
```

src/app/shared/services/switchboard-toaster.service.ts
```diff
@@ -0,0 +1,76 @@
...
+        this.updateMessageList(message, MessageType.success);
+    }
+
+    warning(message?: string, title?: string, override?: Partial<IndividualConfig>): any {
+        this.toaster.warning(message, title, override);
+        this.updateMessageList(message, MessageType.warning);
+    }
+
+    private updateMessageList(message: string, type: string): void {
+        let list = this.massageList.getValue();
```
DawidGil-jlabs, line 69: As with object we could do sth like this here: 
```suggestion
        const list = [{message, type, isNew: true}, ...this.massageList.getValue()];
```
What do you think?

DawidGil-jlabs (commented):
Overall looks good, however I added some comments. Let me know what do you think about them! :)

src/app/layout/header/header.component.html
```diff
@@ -129,49 +129,42 @@ <h4 class="mb-0 f-16 color-white">Tasks</h4>
...
-                    </ng-container>
-
-                    <!-- notif count here -->
-                    <ng-container *ngIf="!isLoadingNotif">
-                        <em class="icon-bell"></em>
-                        <span class="badge badge-danger" *ngIf="notif?.totalCount">{{ notif?.totalCount }}</span>
-                    </ng-container>
-                </a>
+        </li>
+        <li class="nav-item dropdown dropdown-list" dropdown *ngIf="notificationList$ | async as notificationList" (onHidden)="onHiddenNotificationList()">
```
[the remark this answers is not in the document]
VOSpdUkraine: do you think I need to rewrite it in Angular Material?

src/app/layout/header/header.component.scss
```diff
@@ -37,6 +37,13 @@ a:hover {
     bottom: 15px;
 }
 
+.notification-icon {
```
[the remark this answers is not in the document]
VOSpdUkraine: I added some padding, but we need a shield

src/app/layout/header/header.component.html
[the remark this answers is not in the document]
DawidGil-jlabs: No, i didn't say that you should rewrite this. Not now, that's for sure. Just mentioned this problem and asked if you know why is it that :) 
Do you prefer Angular Material or NGX bootstrap?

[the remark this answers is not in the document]
VOSpdUkraine: I prefer Angular Material and don't know why we are using both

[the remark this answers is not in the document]
DawidGil-jlabs: Cool, so let's try stick with Angular Material and if there will be some additional time we will switch existing components to use only Angular Material, sounds good?

src/app/shared/services/switchboard-toastr.service.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+        'error',
+        'info',
+        'warning',
+      ]
+  );
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      providers: [{ provide: ToastrService, useValue: toastrSpy }],
+      imports: [ToastrModule.forRoot()]
```
DawidGil-jlabs, line 22: this import is not needed, because you are mocking already ToastrService. Importing entire module will take more time to build those tests.

src/app/shared/services/switchboard-toastr.service.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    });
+    service = TestBed.inject(SwitchboardToastrService);
+  });
+
+  it('should be created', () => {
+    expect(service).toBeTruthy();
+  });
+
+  it('should be run reset', () => {
+    (service as any).massageList = new BehaviorSubject<SwitchboardToastr[]>([{
```
DawidGil-jlabs, line 32: I think we should test only public methods, not private. 
Private methods should be tested through public ones.

src/app/shared/services/switchboard-toastr.service.ts
```diff
@@ -0,0 +1,70 @@
...
+    message: string;
+    type: string;
+    isNew: boolean;
+}
+@Injectable({
+    providedIn: 'root'
+})
+export class SwitchboardToastrService {
+    private readonly maxLength = 20;
+    private massageList = new BehaviorSubject<SwitchboardToastr[]>([]);
```
DawidGil-jlabs, line 23: I didn't saw it earlier, should be `messageList`

src/app/shared/services/switchboard-toastr.service.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+      isNew: true,
+      message: 'test',
+      type: 'test'
+    }]);
+    spyOn(service, 'reset');
+    service.reset();
+    expect(service.reset).toHaveBeenCalled();
```
DawidGil-jlabs, line 39: I don't understand those tests. You are calling a method and checking if you called it?
I think it would be better if we check that ToastrService methods were called.

src/app/shared/services/switchboard-toastr.service.ts
```diff
@@ -0,0 +1,70 @@
...
+        this.updateMessageList(message, MessageType.success);
+    }
+
+    warning(message?: string, title?: string, override?: Partial<IndividualConfig>): any {
+        this.toastr.warning(message, title, override);
+        this.updateMessageList(message, MessageType.warning);
+    }
+
+    private updateMessageList(message: string, type: string): void {
+        const list = [{message, type, isNew: true}, ...this.massageList.getValue()].slice(0, this.maxLength);
```
DawidGil-jlabs, line 67: Maybe not maxLength but maxMessagesNumber/maxMessagesAmount or something like that ?

src/app/layout/header/header.component.scss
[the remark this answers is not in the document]
DawidGil-jlabs: What do you think about width and height equal to 16px? I think that those icons looks slightly better when they are smaller :)

src/app/shared/services/switchboard-toastr.service.spec.ts
```diff
@@ -18,26 +18,26 @@ fdescribe('SwitchboardToastrService', () => {
...
       type: 'test'
     }]);
     spyOn(service, 'reset');
     service.reset();
     expect(service.reset).toHaveBeenCalled();
-    expect((service as any).massageList.getValue.length).toBe(0);
+    expect((service as any).messageList.getValue.length).toBe(0);
   });
 
   it('should be run getMessageList', () => {
```
DawidGil-jlabs, line 43: I think this test can be deleted.

src/app/shared/services/switchboard-toastr.service.spec.ts
```diff
@@ -47,7 +47,7 @@ fdescribe('SwitchboardToastrService', () => {
   });
 
   it('should be run readAllItems', () => {
```
DawidGil-jlabs, line 49: Maybe something like below?
```suggestion
  it('should be run readAllItems', (done) => {
    service.success('test');
    const expectedResult = {message: 'test', type: 'toast-success', isNew: false};
    service.readAllItems();
    service.getMessageList().subscribe(list => {
      expect(list).toEqual([expectedResult]);
      done();
    });
  });
```

DawidGil-jlabs (commented):
Last two comments that i see after update :)

DawidGil-jlabs (approved):

PR #304, opened 2021-06-17
SWTCH-1045 prepare sb for staking functionality

DawidGil-jlabs:
Creating module and basic routing with component for staking functionality.
I moved also dashboard cards into component, so it will be easier to change/maintain this part of code.

### DIFF
--- a/src/app/layout/header/header.component.html
+++ b/src/app/layout/header/header.component.html
@@ -48,6 +48,14 @@
                     <h6 class="mb-0">Applications</h6>
                 </a>
             </li>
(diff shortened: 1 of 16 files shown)


jrhender:
Does this include any changes that we would need to hide before pushing dev Switchboard to master/volta Switchboard?

DawidGil-jlabs:
Yes, there are changes that shouldn't be visible on master/production environment. I have another task to create feature flag which will hide that functionality. I did this one as first, because I didn't want to block others, and there were doubts if we should implement feature toggle functionality.

VOSpdUkraine (approved):

jrhender (approved):

PR #306, opened 2021-06-17
SWTCH-1049 submit button disabled with valid form

DawidGil-jlabs:
Update info after refactoring and adding tests to enrolmentForm.
1. Removed few not needed properties from `request-claim.component.ts`
2. Extracted to new component entire form for enrolment.
3. Added additional error messages to: datepicker, registration types checkboxes (on/off chain).
4. Added tests for: on/off chain and submit button, displayed error messages (if they appear and have correct message). I think it is useful because we are generating there those fields so it's better to check if everything is fine.

If there are any proposition to test something more or to change some description, feel free to comment :)

### DIFF
--- a/src/app/routes/registration/enrolment-form/enrolment-form.component.html
+++ b/src/app/routes/registration/enrolment-form/enrolment-form.component.html
@@ -1,6 +1,7 @@
 <form role="form" [formGroup]="enrolmentForm">
     <div formGroupName="registrationTypes">
         <mat-checkbox class="pl-3 md-enroll"
+                      data-qa-id="on-chain"
                       formControlName="onChain"
                       [ngStyle]="txtboxColor">
                   <span class="mb-3 d-inline-flex"
@@ -10,6 +11,7 @@
                   </span>
         </mat-checkbox>
         <mat-checkbox class="pl-3 md-enroll"
+                      data-qa-id="off-chain"
                       formControlName="offChain"
                       [ngStyle]="txtboxColor">
                     <span class="mb-3 d-inline-flex"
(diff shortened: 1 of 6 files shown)


jrhender:
Would you be able to add tests for scenarios related to this? For example:
- Request button is disabled when neither checkbox is selected
- Request button is enabled when one of the two checkboxes is selected

DawidGil-jlabs:
> Would you be able to add tests for scenarios related to this? For example:
> 
> * Request button is disabled when neither checkbox is selected
> * Request button is enabled when one of the two checkboxes is selected

To do that i need to mock a lot of things. Because we are setting different things when component is initializing i need to mock/stub: 
1. `activateRoute.queryParams`
2. loading service
3. router
4. methods from iam-client-lib which are: `getDefinition`, `getRolesByNamespace`, `getClaimsByRequester` (and probably `getClaimsBySubject` for my assets), `getUserClaims`, `getDid`, `registrationTypesOfRoles`. And those responses need to have sensible part of data if we want to test those checkboxes.

Problem here was because i didn't cleaned up correctly enrolmentForm. However we have tests about making invalid form when none of the selected checkboxes are selected. [Tests for required at least one checkbox](https://github.com/energywebfoundation/switchboard-dapp/blob/develop/src/app/utils/validators/require-checboxes-to-be-checked.validator.spec.ts).

Anyway it's up to you, i can make some tests and not only for submit button, but it will take some time to setup everything.

jrhender (commented):
Based on discussion @DawidGil-jlabs will refactor a bit to add more tests. @DawidGil-jlabs Please let me know when ready for review again 🙇

jrhender (commented):
Noticing an issue where even if a role is eligible on-chain, the checkbox is disabled. I think it might be because `this.namespaceRegistrationRoles` is undefined at the time that `updateEnrolmentForm` is called

image

DawidGil-jlabs:
Ok, it's now fixed (it worked for me :D ). There was an easy fix for it, just to set registration roles before field list.

jrhender (approved):
Great, looking good!

PR #308, opened 2021-06-21
[SWTCH-1000] Staking pool form

davycastillo:


src/app/routes/applications/governance-list/governance-list.component.html
```diff
@@ -172,6 +172,12 @@
                 <span>Create Role</span>
               </button>
 
+              <!-- Create Staking Pool -->
+              <button mat-menu-item class="btn-color-primary" (click)="openStakingPool(element)">
```
DawidGil-jlabs, line 176: Please add `*appIsFeatureEnabled` here. It works like ngIf, but listens on feature flag. So it will be visible on switchboard-dev but not on volta (prod environment). I think you will need to rebase/merge changes from develop branch.

src/assets/img/icons/create-staking-icon.svg
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
```
DawidGil-jlabs, line 1: Would be awesome if you minify this svg for example with https://jakearchibald.github.io/svgomg/ :)

PR #309, opened 2021-06-22
SWTCH-1046 feature flag for staking functionality

DawidGil-jlabs:
I created a directive that works like ngIf, but looks on property featureVisible in environment.
There is also a guard that blocks navigating to specific route. So there will be no possibility to navigate to a disabled feature, because it will be blocked.
I did it generic at this moment.


VOSpdUkraine (approved):

jrhender (approved):
Looks good to me.

PR #311, opened 2021-06-23
[SWTCH-913] Switchboard to call OrgCreator service

VOSpdUkraine:


src/app/routes/applications/applications.component.ts
```diff
@@ -65,6 +65,13 @@ export class ApplicationsComponent implements OnInit, AfterViewInit, OnDestroy {
   }
 
   openNewOrgComponent(): void {
+    if (!this.isIamEwcOwner) {
+      const namespace = 'orgowner.roles.orgcreator.apps.energyweb.iam.ewc';
```
jrhender, line 69: I think `namespace` should just be the app name? Not sure about that though, you would need to test to confirm
```suggestion
      const namespace = 'orgcreator.apps.energyweb.iam.ewc';
```

src/app/routes/applications/applications.component.ts
```diff
@@ -65,6 +65,13 @@ export class ApplicationsComponent implements OnInit, AfterViewInit, OnDestroy {
...
+      const roleName = 'orgowner';
+      const siteUrl = `${location.origin}/#/enrol?org=${namespace}&roleName=${roleName}`;
```
jrhender, line 71: Url should look similar to `https://switchboard-dev.energyweb.org/#/enrol?roleName=orgowner&app=orgcreator.apps.testorg.iam.ewc`. This is because `orgcreator` is an `app` not an `org`
```suggestion
      const siteUrl = `${location.origin}/#/enrol?app=${namespace}&roleName=${roleName}`;
```

src/app/routes/applications/applications.component.ts
```diff
@@ -65,6 +65,13 @@ export class ApplicationsComponent implements OnInit, AfterViewInit, OnDestroy {
...
+      window.open('//' + siteUrl, '_blank');
```
jrhender, line 72: Instead of `window.open`, could maybe just route to enrolment page. This is what is done when the `enrol` button is clicked from search screen. @VOSpdUkraine please adjust based on feedback from Olesia and John M in story. IMO, it would be better to just navigate to enrolment screen rather than new window but interested in your, Olesia and John M's opinions.
image

jrhender (commented):
Good start, added some comments. Need feedback on UX from Olesia and John M

jrhender (approved):
Looks good but **please wait to merge** until the org creator service is deployed and available. Otherwise the functionality will be confusing to users

PR #314, opened 2021-06-24
Fix/swtch 1010 add issuer of role

VOSpdUkraine:


DawidGil-jlabs (approved):
Looks good, but could you fix also a problem with validation? There is an error message when i change issuer type to role, but i can proceed. This make some confusion to user, because he get info that this form is invalid. Or you want to do that in another PR?

VOSpdUkraine:
Hi @DawidGil-jlabs 
about "required"  ok
but what about  minLength(3), maxLength(256), isAlphanumericValidator ?

image

DawidGil-jlabs:
Well, if that's not a problem you can copy paste from new Role Name validation error message for minlength: 
image And check how it works. About `isAlphanumericValidator`: for issuer role namespace `.` is allowed, but not for Role Name (look on method `(keypress)="alphaNumericOnly($event)"`). This is a mess for me, because for one control we have two different behaviors.

I see two solutions, but it doesn't need to be done in this PR imo:
1. Split role name and issuer role namespace into two separate properties in form.
2. Add dynamically validators depending on issuer type.

maxlength validator doesn't need an error message, because you can't write more than 256 characters. Maybe we should remove it? 🤔 

What do you think?

PR #315, opened 2021-06-24
[SWTCH-1086] Fix show multi signature messages

VOSpdUkraine:


jrhender (commented):
@VOSpdUkraine changes look good, I will test to confirm. How difficult would it be to add unit tests for this component? Do you think it would be worthwhile in this case? I suppose we need to mock a few of the injected services but it seems doable at first glance, IMO.

@DawidGil-jlabs Do you think tests of the `view-requests.component` are worthwhile?

jrhender (changes requested):
@VOSpdUkraine Still one case to cover please. The message should not show when the claim request is approved, even for issuers: 
image

DawidGil-jlabs:
@jrhender about testing. With @VOSpdUkraine we agreed that it would be good to write unit tests to every new code, at least there where is logic. Existing components are hard to test because they contain to many dependencies.

However this component isn't complicated and there is only one method from IamService that need to be mocked (in ngOnInit). Of course there is a need to mock every service, but most of them are mocked in different tests so this can be done by copy pasting :) I think test would be cool here, because this logic already made some problems.

Anyway this would be a lot easier to test if all logic will be in store (effects/reducers/selectors) by creating an enrolment store slice. I'm not sure if there is time for that, because few days left to the end of the sprint and we have pending tasks.

DawidGil-jlabs:
Additional info about testing. It could be also pretty easy if we add a service and move there communication with backend. So that service would have those dependencies: LoadingService, IamService, NotificationService, SwitchboardToastrService. There will be less mocking in component that way it might be easier to test component and service.

jrhender (approved):
Looks good!

PR #316, opened 2021-06-28
SWTCH-1035 list of methods doesnt change

DawidGil-jlabs:
Adds retry when updating verification methods list. It tries 5 times every 2 seconds. If list will be the same after retries it will display a message that backend doesn't have newest blockchain data.
In my tests I saw that max time difference was 3-4 seconds.


jrhender (commented):
Looks good, thanks!

jrhender (approved):

PR #317, opened 2021-06-28
SWTCH 1085 UI implementation

amaugosomto:
1. Added the feature to make it possible to edit an already added field when adding a role.
2. Fixed the bug of date not showing on date field card.


src/app/routes/applications/new-role/new-role.component.ts
```diff
@@ -393,6 +394,16 @@ export class NewRoleComponent implements OnInit, AfterViewInit, OnDestroy {
     }
   }
 
+  updateField() {
+    if (this.fieldsForm.valid) {
+      this.dataSource.data.splice(this.fieldIndex, 1, this._extractValidationObject(this.fieldsForm.value))
+      this.fieldsForm.reset();
+      this.showFieldsForm = false;
```
DawidGil-jlabs, line 401: It's more like a question, but is it possible to merge `showFieldsForm` and `isEditFieldForm` into one property? Do you know that?
I'm asking because you are setting both with the same values :)

src/app/routes/applications/new-role/new-role.component.ts
```diff
@@ -465,6 +476,23 @@ export class NewRoleComponent implements OnInit, AfterViewInit, OnDestroy {
     this.dataSource.data = [...list];
   }
 
+  editField(i: number) {
+    this.fieldIndex = i;
+    let field = this.dataSource.data[i];
```
DawidGil-jlabs, line 481: It's a small thing but use const where possible instead of let. We didn't set up tslint/eslint in travis, so this needs to be handled manually :(

src/app/routes/applications/new-role/new-role.component.ts
```diff
@@ -465,6 +476,23 @@ export class NewRoleComponent implements OnInit, AfterViewInit, OnDestroy {
...
+    let fieldKeys = Object.keys(field);
+    let valueToPatch = {}
```
DawidGil-jlabs, line 483: missing semicolon ;)

src/app/routes/applications/new-role/new-role.component.ts
amaugosomto: So `showFieldsForm` is a property that shows or hides the form and `isEditFieldForm` is what toggles the button from save to update

DawidGil-jlabs: Ok, thanks 👍

DawidGil-jlabs:
Like i said on slack, fields in `confirm details` step are not updated when editing fields in `Set fields` step.

davycastillo:
LGTM now. @DawidGil-jlabs ?

DawidGil-jlabs (approved):

PR #320, opened 2021-07-01
SWTCH-988 add bitcoin option to verification method

DawidGil-jlabs:


src/app/routes/assets/pipes/type-algorithm.pipe.ts
```diff
@@ -10,8 +10,14 @@ export class TypeAlgorithmPipe implements PipeTransform {
...
       return '';
     }
-    const index = this.getIndex(value);
-    return index < 0 ? value : this.getKeys(index);
+
+    const result = Object.entries(AlgorithmsEnum).filter((entry) => {
+      if (value.includes(entry[1])) {
+        return entry[0];
+      }
+    }).map(entry => entry[0]).join(' / ');
```
jrhender, line 18: Would consider encapsulating this logic into function so that you can give a name to this block of code to make it easier to read. Maybe a matter of preference though.
```suggestion
    const getFriendlyNamesForKeyType = (keyType: string): string => Object.entries(AlgorithmsEnum).filter((entry) => {
      if (keyType.includes(entry[1])) {
        return entry[0];
      }
    }).map(entry => entry[0]).join(' / ');
    const result = getFriendlyNamesForKeyType(value);
```

src/app/routes/assets/models/algorithms.enum.ts
```diff
@@ -1,3 +1,4 @@
 export enum AlgorithmsEnum {
```
jrhender, line 1: I know it was already like this, but would consider renaming this to `KeyTypesEnum` as I think it is more accurate than `AlgorithmsEnum`
```suggestion
export enum KeyTypesEnum {
```

jrhender (approved):
A couple suggestions but looks good overall 👍

src/app/routes/assets/pipes/type-algorithm.pipe.ts
DawidGil-jlabs: Yeah, i didn't cleaned it up. I don't know why I forgot about it. So I removed not used methods there and extracted that part to a method.

PR #321, opened 2021-07-02
SWTCH-1101 missing error messages

DawidGil-jlabs:
There was no error messages, only highlighted fields. So i added them, I wrote few unit tests and refactored a little bit.


jrhender (approved):
Looks good to me

PR #335, opened 2021-07-16
SWTCH 994 multiple key validation

amaugosomto:


src/app/routes/assets/verification-method/verification-method.component.ts
```diff
@@ -55,6 +57,15 @@ export class VerificationMethodComponent implements OnInit {
     if (this.isFormDisabled) {
       return;
     }
+
+    const publicKeyExists = this.dataSource.map(key => key.publicKeyHex === this.publicKey.value);
```
DawidGil-jlabs, line 61: What do you think about creating a validator that will check if putted value do not exist in list?

src/app/routes/assets/verification-method/verification-method.component.ts
```diff
@@ -30,7 +31,8 @@ export class VerificationMethodComponent implements OnInit {
 
   constructor(private dialogRef: MatDialogRef<VerificationMethodComponent>,
               @Inject(MAT_DIALOG_DATA) private dialogData: any,
-              private verificationService: VerificationService) {
+              private verificationService: VerificationService,
+              private toastr: SwitchboardToastrService) {
```
DawidGil-jlabs, line 35: By adding toastr you need to add provider in test. That way your tests stop failing.

src/app/routes/assets/verification-method/verification-method.component.ts
```diff
@@ -55,6 +57,15 @@ export class VerificationMethodComponent implements OnInit {
     if (this.isFormDisabled) {
       return;
     }
+
+    const publicKeyExists = this.dataSource.map(key => key.publicKeyHex === this.publicKey.value);
+    if (publicKeyExists) {
+      
+      this.toastr.error('Public key entered already exists, please choose another');
```
DawidGil-jlabs, line 64: This `toastr` or i should say `SwitchboardToastrService` is adding notification in header. What i mean is in screenshot. I'm not sure if this is required to have such notifications here. Maybe we should ask Olesia?
image
@VOSpdUkraine what do you think?

DawidGil-jlabs (commented):
I added few comments, because I'm not sure if adding notification is the best option.

src/app/routes/assets/verification-method/verification-method.component.ts
amaugosomto: Okay. You mean in a different function right?

DawidGil-jlabs: Yeah. I think validator is fine here, but you might need to create async validator. I'm not sure, never done something like this 😛

amaugosomto: I do not understand what you mean by async validator

DawidGil-jlabs: This: https://angular.io/api/forms/AsyncValidator 
But you can fix it without that:
1. Create pure function to validate that -> returns true/false.
2. assign value returned from that function to some property.
3. display message if value is duplicated.
4. block sending form if input value is wrong.
5. Adjust styles.
6. Maybe something more to do.

amaugosomto:
Had to close PR and delete the branch as branch codes are too different from origin/develop and a merge modifies up to 89 files which would be too hard to refactor.

PR #336, opened 2021-07-16
SWTCH-1111 create patron page for staking functionality.

DawidGil-jlabs:
Page for staking.
To use it by going to `http://localhost:4200/#/staking?org=dawidgil.iam.ewc` or `dmitryfesenko.iam.ewc`, or any other organization that is a provider.
There is a need to add more unit tests and some changes, but it would be good to have it already for testing purposes on dev environment.


jrhender (approved):
Looks good 👍

PR #337, opened 2021-07-16
fix(patron-login): initCacheServer&initDID = false

jrhender:
only metamask connection is necessary for patron staking
No signature should be required

rename to `initDID` to match `iam-client-lib`


DawidGil-jlabs (approved):
Lgtm. 
Sorry for this. I thought it is set to false. Maybe when i was testing something i changed it into true. ;/

PR #339, opened 2021-07-19
SWTCH-1111 change way of login for staking

DawidGil-jlabs:
1. Changes login 
 -> now user can login by clicking a button in top right corner to login
 -> put value to stake and click stake button. After login it should try to putStake.
   - if value is bigger than balance, then balance will be used instead.
   - if user already put a stake to this provider, then nothing will be done. Just a message, that user already has put stake.
There is a workaround for handling stake after login. And there is no tests right now for it.
2. Removes lazy loaded module for staking, checking if this makes a problem with azure.
3. When user changes an account in metamask then he will no longer be redirected to main switchboard page. Now it should disconnect and reload a page.
4. Added error messages for stake amount input.
5. Refresh the page when user logout.

Things that are needed to be done in next PRs: Unit tests for every part of this view.


jrhender (approved):
Seems good to me

PR #340, opened 2021-07-20
SWTCH-1121 remove hash from url

DawidGil-jlabs:
In this PR I'm removing `#` from url. There was less work than i thought.


jrhender (approved):
Nice! a small change after all!

VOSpdUkraine (approved):

PR #345, opened 2021-07-26
fix(SWTCH-1148) change chain config 

DawidGil-jlabs:


JGiter (commented):
Looks good. But please wait for https://github.com/energywebfoundation/iam-client-lib/pull/219 is merged and then update iam-lib, because previously deployed pool with normal reward growing didn't expose withdraw delay and minimum staking interval. This PR also add some more changes
-  `putStake`, `requestWithdraw` and `withdraw` receives optional second parameter - desired transaction speed
- `withdrawalDelay` and `requestWithdrawalDelay` returns number

JGiter (approved):

PR #349, opened 2021-07-26
Swtch 963 role creation should

amaugosomto:


VOSpdUkraine (approved):

DawidGil-jlabs (approved):
Looks good. Just reuse one interface instead of creating two and i think you can merge it :)

PR #350, opened 2021-07-27
finished chore of multiple key validation

amaugosomto:
I deleted the old branch and created this new one as trying to merge made changes to 89 files. Also removed the previous toastr error implementation and used a form error.

### DIFF
--- a/src/app/routes/assets/verification-method/verification-method.component.ts
+++ b/src/app/routes/assets/verification-method/verification-method.component.ts
@@ -5,6 +5,7 @@ import { AbstractControl, FormControl, ValidationErrors, ValidatorFn, Validators
 import { PageEvent } from '@angular/material/paginator';
 import { KeyTypesEnum } from '../models/keyTypesEnum';
 import { isHexValidator } from '../../../utils/validators/is-hex.validator';
+import { listContainsValidator } from '../../../utils/validators/list-contains.validator';
 
 export interface PublicKey {
   publicKeyHex: string;
@@ -23,15 +24,7 @@ export class VerificationMethodComponent implements OnInit {
   dataSource: PublicKey[] = [];
   selectControl = new FormControl('', [Validators.required]);
 
-  publicKeyExistValidator(): ValidatorFn {
-    return (control: AbstractControl): ValidationErrors | null => {
-      const exists = this.dataSource.find(key => key.publicKeyHex === control.value);
-      return exists ? {publicKeyExist: {value: control.value}} : null;
-    };
-  }
-  publicKey = new FormControl('',
-    [Validators.required, isHexValidator, this.publicKeyExistValidator()]
-  );
+  publicKey = new FormControl('');
   selectOptions = Object.entries(KeyTypesEnum);
   private publicKeys;
 
@@ -79,7 +72,7 @@ export class VerificationMethodComponent implements OnInit {
       return 'Invalid input. Public key must start with "0x" to be followed by 66 or 130 Hexadecimal characters.';
     }
 
-    if (this.publicKey.hasError('publicKeyExist')) {
+    if (this.publicKey.hasError('listContains')) {
       return 'Public key entered already exists, please choose another.';
     }
 
@@ -95,10 +88,12 @@ export class VerificationMethodComponent implements OnInit {
     this.publicKeys = publicKeys;
     this.verificationsAmount = publicKeys.length;
     this.setDataSource();
+    this.publicKey.setValidators([Validators.required, isHexValidator, listContainsValidator(this.dataSource, 'publicKeyHex')]);
   }
 
   private setDataSource(): void {
     this.dataSource = this.publicKeys.slice(this.pageIndex * this.pageSize, (this.pageIndex + 1) * this.pageSize);
+    console.log(this.dataSource)
   }
 
   private loadPublicKeys(): void {
--- a/src/app/routes/assets/verification-method/verification-method.component.ts
+++ b/src/app/routes/assets/verification-method/verification-method.component.ts
@@ -1,7 +1,7 @@
 import { Component, Inject, OnInit } from '@angular/core';
 import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material/dialog';
 import { VerificationService } from './verification.service';
-import { AbstractControl, FormControl, ValidationErrors, ValidatorFn, Validators } from '@angular/forms';
+import { FormControl, Validators } from '@angular/forms';
 import { PageEvent } from '@angular/material/paginator';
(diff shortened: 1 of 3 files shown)


src/app/routes/assets/verification-method/verification-method.component.ts
```diff
@@ -22,8 +22,15 @@ export class VerificationMethodComponent implements OnInit {
   verificationsAmount: number;
   dataSource: PublicKey[] = [];
   selectControl = new FormControl('', [Validators.required]);
+
+  publicKeyExistValidator(): ValidatorFn {
```
DawidGil-jlabs, line 26: Method shouldn't be mixed with properties. Property > constructor > methods. Please align that.

src/app/routes/assets/verification-method/verification-method.component.ts
```diff
@@ -22,8 +22,15 @@ export class VerificationMethodComponent implements OnInit {
...
+    return (control: AbstractControl): ValidationErrors | null => {
```
DawidGil-jlabs, line 27:
```suggestion
      listContainsValidator(source: unknown[], property: string): ValidatorFn {
    return (control: AbstractControl): ValidationErrors | null => {
      const exists = source.find(key => key[property] === control.value);
      return exists ? {listContains: {value: control.value}} : null;
    };
  }
```
What do you think? This way we can remove this validator from component, writing unit tests will be pretty easy.
The only thing is that you need to set validators later like this: `this.publicKey.setValidators([Validators.required, isHexValidator, this.publicKeyExistValidator(this.dataSource, 'publicKeyHex')]);`, not when initializing. For example in handleLoadedPublicKeys.

@VOSpdUkraine please add your thoughts on this.

DawidGil-jlabs (commented):
Generally looks good. I added 1 comment to align. Second is for discussion.

src/app/routes/assets/verification-method/verification-method.component.ts
amaugosomto: Yeah. I was thinking of putting the function in utils/validators folder. But I didnt see any other component using it. So I left it there

DawidGil-jlabs: Ok, it would be good to have tests for it imo.
I think having it as a function will make testing a lot easier. -> you do not have to mock component properties. However it's your call, both solutions are fine for me.

src/app/routes/assets/verification-method/verification-method.component.ts
```diff
@@ -83,10 +88,12 @@ export class VerificationMethodComponent implements OnInit {
     this.publicKeys = publicKeys;
     this.verificationsAmount = publicKeys.length;
     this.setDataSource();
+    this.publicKey.setValidators([Validators.required, isHexValidator, listContainsValidator(this.dataSource, 'publicKeyHex')]);
   }
 
   private setDataSource(): void {
     this.dataSource = this.publicKeys.slice(this.pageIndex * this.pageSize, (this.pageIndex + 1) * this.pageSize);
+    console.log(this.dataSource)
```
VOSpdUkraine, line 96: is it need ?

src/app/utils/validators/validators.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { FormControl } from '@angular/forms';
+import { listContainsValidator } from './list-contains.validator';
+
+describe('Validators', () => {
```
DawidGil-jlabs, line 4: This describe can be removed.

src/app/routes/assets/verification-method/verification-method.component.ts
amaugosomto: forgot to remove it. Done and pushed

src/app/utils/validators/validators.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+    describe('listContainsValidators', () => {
+        it('listContainsValidator should not return null if it fails', () => {
+            const source = [
+                { publicKeyHex: '0x8C2CBcBE93b96ef66F854407548edE5f74f4c016ybdybdjdbdjdyydydjdjdjdjdd' }
```
DawidGil-jlabs, line 8: You can leave this as it is, but do you know you can put here whatever you want? It doesn't need to be publicKeyHex.

src/app/utils/validators/validators.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+        });
+
+        it('listContainsValidator should return null if it passes', () => {
+            
+            const validator = listContainsValidator([], 'publicKeyHex');
+            const exists = validator(new FormControl('0x8C2CBcBE93b96ef66F854407548edE5f74f4c016ybdybdjdbdjdyydydjdjdjdjdd'));
+            
+            expect(exists).toEqual(null);
+        });
+    })
```
DawidGil-jlabs, line 24: What would happen if you put undefined instead of empty array in source? 
What would happen if you put undefined as a property?
Do you think those a valid cases?

DawidGil-jlabs (approved):

src/app/routes/assets/verification-method/verification-method.component.ts
```diff
@@ -95,10 +88,12 @@ export class VerificationMethodComponent implements OnInit {
...
```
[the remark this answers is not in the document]
amaugosomto: removed

src/app/utils/validators/validators.spec.ts
amaugosomto: I put it like that in case the other validators need to be tested

src/app/utils/validators/validators.spec.ts
amaugosomto: Yeah, I know. I felt it would be better to test with what we are already using. I could use other keys

DawidGil-jlabs: But we do not have any wrapper for those validators. There are few our own validators and each have it's own file with testing. We can in fact create a class for example:
```
  class SbValidators {
    static listContains() {}
    static someOtherMethod
    ....
  }
```
That way this describe wrapper have more sense. But this is topic for a different discussion.

src/app/utils/validators/validators.spec.ts
DawidGil-jlabs: But we do not have any wrapper for those validators. There are few our own validators and each have it's own file with testing. We can in fact create a class for example:
```
  class SbValidators {
    static listContains() {}
    static someOtherMethod
    ....
  }
```
That way this describe wrapper have more sense. But this is topic for a different discussion.

DawidGil-jlabs (approved):
Lgtm

PR #353, opened 2021-07-28
Fixed Issue with year on the role restrictions step

VOSpdUkraine:
#SWTCH-1152

### DIFF
--- a/src/app/routes/applications/new-role/components/role-field/role-field.component.html
+++ b/src/app/routes/applications/new-role/components/role-field/role-field.component.html
@@ -92,18 +92,18 @@
                 <mat-label class="ml-3 pl-1">Min Date</mat-label>
                 <mat-form-field class="mt-2" appearance="outline" floatPlaceholder="never">
                   <input [matDatepicker]="pickerMin" matInput placeholder="Min Date" autocomplete="off"
-                         formControlName="minDate" (keypress)="$event.preventDefault()">
+                         formControlName="minDate" disabled>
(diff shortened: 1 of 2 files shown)


src/app/routes/applications/new-role/components/role-field/role-field.component.html
```diff
@@ -92,18 +92,18 @@
                 <mat-label class="ml-3 pl-1">Min Date</mat-label>
                 <mat-form-field class="mt-2" appearance="outline" floatPlaceholder="never">
                   <input [matDatepicker]="pickerMin" matInput placeholder="Min Date" autocomplete="off"
-                         formControlName="minDate" (keypress)="$event.preventDefault()">
+                         formControlName="minDate" disabled>
```
DawidGil-jlabs, line 95: I'm not a designer so i'll ask our expert. @davycastillo what do you think from UX perspective about disabling ability to write a date by user? It is desired behavior?

davycastillo: I think it's more convenient to allow the user to edit the date but, just need to add the error message accordingly.

VOSpdUkraine: if you do this, then you need to add more than just an error message to a full-fledged validation with valid characters and formats

dawidgil (approved):
Like we said on standup meeting. Let's go with this approach for now.

PR #364, opened 2021-08-04
fix enrolment navigation.

dawidgil:
Problem occurs when trying to navigate from `enrolment requests` to `my enrolments` in `Enrolments` page. Previously it worked because we had loose equality. Now we have strict equality and we are comparing string to number.


davycastillo (approved):
LGTM.

PR #366, opened 2021-08-05
SWTCH-1142 implement button to copy staking url

dawidgil:
1. Checks if organization is a provider - if not shows action to create staking pool.
2. Move actions to separate components for org/apps/roles.
3. Show copy to clipboard staking url only when organization is provider. 
4. Removes double checking if an organization/role/app belongs to the current user.


davycastillo (approved):
LGTM.

PR #368, opened 2021-08-05
fix switchboard header logo navigation.

dawidgil:
When clicking on logo, page is reloaded and uses #. This fix will stop refreshing a page and removes # from url.


davycastillo (approved):
LGTM

PR #373, opened 2021-08-09
 added the next improvements

VOSpdUkraine:


src/app/app.module.ts
```diff
@@ -81,6 +82,7 @@ if (environment.SENTRY_DNS) {
     StoreModule.forRoot(rootReducer, {}),
     StoreDevtoolsModule.instrument({maxAge: 25, logOnly: environment.production}),
     EffectsModule.forRoot([UserEffects, StakeEffects, AuthEffects]),
+    NgxEditorModule
```
dawidgil, line 85: Adding here this module is not needed afaik. It should work without it here.

src/app/routes/applications/new-staking-pool/new-staking-pool.component.scss
```diff
@@ -42,6 +44,97 @@
...
+
+    .NgxEditor {
+      border: none;
+      border-radius: 0;
+      background: #1C1726;
+      color: white;
+      height: $editor-height;
+      overflow-y: scroll;
+
+      &__Popup {
```
dawidgil, line 63: Do you like styling this way?

dawidgil (commented):
I think it's not finished yet. There is no function for handling this editor data. I know there is no support from iam-lib but it's possible to handle this on frontend side. So when iam-lib enable setting terms and conditions on their side, we can do that also when updating library.

src/app/routes/applications/new-staking-pool/new-staking-pool.component.scss
davycastillo: maybe instead of showing the URL pop up below. It will show on top?

davycastillo (approved):

davycastillo (changes requested):
@dawidgil styling is OK on my side.

VOSpdUkraine:
this editor returns the string

src/app/layout/header/dialog-user/dialog-user.component.spec.ts
```diff
@@ -105,8 +105,7 @@ describe('DialogUserComponent', () => {
 
     fixture.detectChanges();
     const matError = hostDebug.query(By.css('mat-error')).nativeElement;
-
-    expect(matError.textContent).toContain('Maximum allowed date is');
+    expect(matError.textContent).toContain('Required');
```
dawidgil, line 108: No, no no. If you do not know why test is failing, do not change it to "passing" state, because it is useless right now. There is a problem and you need to debug it. Not just changing to have everything green!
I checked what happened to this test and the solution is by changing function `todayDate` at the bottom of file: 
```
    const todayDate = () => {
  const date = new Date(Date.now());
  return `${date.getMonth()}/${date.getDate()}/${date.getFullYear()}`;
};
```
I just made a mistake there with creating a date, nothing more.

dawidgil: fix is here: https://github.com/energywebfoundation/switchboard-dapp/pull/379

src/app/layout/header/dialog-user/dialog-user.component.html
```diff
@@ -27,7 +27,7 @@ <h4 mat-dialog-title class="mb-0 d-flex justify-content-between align-items-cent
             <mat-datepicker-toggle matSuffix [for]="picker"></mat-datepicker-toggle>
             <mat-datepicker #picker></mat-datepicker>
             <mat-error *ngIf="profileForm.get('birthdate').hasError('required')">Date of birth is
-              <strong>required</strong>
+              <strong>Required</strong>
```
dawidgil, line 30: Why changing this?

src/app/routes/applications/new-staking-pool/new-staking-pool.component.ts
```diff
@@ -17,8 +18,8 @@ const {parseEther} = utils;
   templateUrl: './new-staking-pool.component.html',
   styleUrls: ['./new-staking-pool.component.scss']
 })
-export class NewStakingPoolComponent implements OnInit {
-
+export class NewStakingPoolComponent implements OnInit, OnDestroy {
+  editor: Editor;
```
dawidgil, line 22: Please add handling returned value of this editor inside form - it is supported as far as i see from docs. That way we can add validators if that will be needed and we will have only one check for enabling/disabling submit button.

PR #375, opened 2021-08-10
SWTCH-1105 handle stake ewt pages

dawidgil:
Now going to stake-ewt page you will see a real list of providers instead of mock. Clicking on them will navigate to generic staking patron page in new tab.


davycastillo (approved):
Looks good

PR #377, opened 2021-08-12
 added the next improvements

VOSpdUkraine:


src/app/routes/applications/new-role/new-role.component.html
```diff
@@ -41,6 +41,7 @@ <h4 mat-dialog-title class="mb-0 d-flex justify-content-between align-items-cent
                       alphanumeric characters.
                     </mat-error>
                   </mat-form-field>
+                    <mat-error *ngIf="isExistsRoleName">This mane already exists. Please try another</mat-error>
```
dawidgil, line 44: Should be `This name...`

src/app/routes/applications/new-role/new-role.component.ts
```diff
@@ -89,7 +89,7 @@ export class NewRoleComponent implements OnInit, AfterViewInit, OnDestroy {
   });
 
   public roleControl = this.fb.control('');
-  public restrictionRoleControl = this.fb.control('', [Validators.required, this.roleNotFoundValidator(), this.roleExistValidator()]);
```
dawidgil, line 92: Implementation of those validators is used somewhere? If not, please remove it.

src/app/utils/functions/is-alpha-numeric.ts
```diff
@@ -6,6 +6,6 @@ export const isAlphaNumericOnly = (event: any, includeDot?: boolean): boolean =>
     (charCode > 96 && charCode < 123) || // a-z
     (charCode > 64 && charCode < 91) || // A-Z
     (charCode > 47 && charCode < 58) || // 0-9
-    (includeDot && charCode === 46) // .
+    (!!includeDot && charCode === 46) // .
```
dawidgil, line 9: why? It is already a boolean value.

dawidgil (commented):
Updating list of preconditions doesn't work. When you go to role governance tab -> edit role -> try to set preconditions. It didn't work for me.

src/app/utils/functions/is-alpha-numeric.ts
VOSpdUkraine: event "keypress" need only true or false if we return undefined it is not work correct

dawidgil: Oh, ok. That's why we need tests for such cases...

VOSpdUkraine:
@dawidgil please add more details what is not work ?

dawidgil:
I was editing a role, and wanted to add some precondition to edited role. I was selecting a role, and after clicking `+` sign it didn't showed up on the list. I also got some errors in console. I suppose setting precondition is not connected with form when editing role.
Anyway are you able to add precondition to existing role?

dawidgil (commented):
When you create a role with only one precondition, or you edit a role without precondition and add only one precondition while editing it's not working correctly.
Steps to reproduce: go to role without precondition -> add one role -> edit -> after successful edition go again to that role and edit it -> you will see empty element as a precondition like on the screen:
image

PR #378, opened 2021-08-12
Clean_up_list_of_dependencies

VOSpdUkraine:


dawidgil (approved):
What about:
- fullcalendar
- jquery
- lodash
- popper
? Did you check them?

VOSpdUkraine:
What about:

fullcalendar
jquery
lodash
popper
? Did you check them?

yes i checked them they are used in different places of our project

dawidgil:
Strange, because lodash as far as i remember was used in directive which wasn't used in application. In other words we could remove directive and then remove lodash. Did you check it this way?

VOSpdUkraine:
image

dawidgil:
Ok, so i dig a little bit and i see that sidebar is not used in application. It's commented out, so after a quick talk with @davycastillo i think we can remove also sidebar. 
I think we should remove dead code.

davycastillo (approved):

PR #382, opened 2021-08-20
SWTCH-1164 improve login functionality

dawidgil:
Rewriting and unifying authorization. Few parts are divided + few tests. Solves few problems from user perspective.


davycastillo (approved):
LGTM

PR #384, opened 2021-08-27
SWTCH-1191 fix: error when getting claims by empty list of subjects.

dawidgil:
Error occurs when we do not have assets at all and we go to assets offered to me.

### DIFF
--- a/src/app/routes/assets/asset-list/asset-list.component.ts
+++ b/src/app/routes/assets/asset-list/asset-list.component.ts
@@ -275,7 +275,7 @@ export class AssetListComponent implements OnInit, OnDestroy {
   private mapEnrolments() {
     return (source: Observable<Asset[]>) => {
       return source.pipe(
-        filter(assets => assets.length > 0),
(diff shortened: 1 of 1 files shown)


jrhender (approved):

src/app/routes/assets/asset-list/asset-list.component.ts
```diff
@@ -275,6 +275,7 @@ export class AssetListComponent implements OnInit, OnDestroy {
   private mapEnrolments() {
     return (source: Observable<Asset[]>) => {
       return source.pipe(
+        filter(assets => assets.length > 0),
```
jrhender, line 278: Is there a chance of a null exception here?
```suggestion
        filter(assets => assets?.length > 0),
```

dawidgil: I don't think so, because we always get here a list. Empty or with assets, but i can add elvis operator as you suggested.

PR #386, opened 2021-08-30
SWTCH-1193 remove setup errors from unit tests

dawidgil:
Removes errors from tests about for example not found selector etc.


jrhender (approved):
Looks good

PR #387, opened 2021-08-30
SWTCH-1188 not reloading page when gets 500 status code error

dawidgil:
This one is not easy to test in switchboard, because it depends on token expiration.

### DIFF
--- a/src/app/shared/services/login/login.service.ts
+++ b/src/app/shared/services/login/login.service.ts
@@ -64,7 +64,7 @@ export class LoginService {
           }
           return Boolean(loginSuccessful);
         }),
-        catchError(err => this.handleLoginErrors(err))
+        catchError(err => this.handleLoginErrors(err, redirectOnChange))
       );
   }
 
@@ -128,7 +128,12 @@ export class LoginService {
   }
 
   private openSwal(config, navigateOnTimeout: boolean) {
-    from(SWAL(config))
+    from(SWAL({
+      icon: 'error',
+      button: 'Proceed',
+      closeOnClickOutside: false,
+      ...config
+    }))
       .pipe(
         take(1),
         filter(Boolean)
@@ -138,11 +143,19 @@ export class LoginService {
       );
   }
 
-  private handleLoginErrors(e) {
+  private handleLoginErrors(e, navigateOnTimeout) {
     console.error(e);
+    console.log(e);
+    debugger;
     const loginError = LOGIN_ERRORS.filter(error => e.message.includes(error.key))[0];
     if (loginError?.type === 'swal') {
-      this.displayLoginErrorWithSwal(loginError.value);
+      this.displayLoginErrorWithSwal(loginError.value, navigateOnTimeout);
+    } else if (e.message === 'Request failed with status code 401') {
+      const config = {
+        title: 'Session Expired',
+        text: 'Please proceed to login again',
+      };
+      this.openSwal(config, navigateOnTimeout);
     } else {
       const message = loginError ? loginError.value : e.message;
       this.toastr.error(message);
@@ -150,16 +163,13 @@ export class LoginService {
     return of(false);
   }
 
-  private displayLoginErrorWithSwal(message: string) {
+  private displayLoginErrorWithSwal(message: string, navigateOnTimeout: boolean) {
     const config = {
       title: 'Wrong Network',
       text: `${message}`,
-      icon: 'error',
-      button: 'Proceed',
-      closeOnClickOutside: false
     };
 
-    this.openSwal(config, true);
+    this.openSwal(config, navigateOnTimeout);
   }
 
   private saveDeepLink(): void {
--- a/src/app/shared/services/login/login.service.ts
+++ b/src/app/shared/services/login/login.service.ts
@@ -146,11 +146,11 @@ export class LoginService {
    */
   private handleLoginErrors(e, navigateOnTimeout) {
     console.error(e);
-    const config = swalLoginError(e.message);
-    if (config) {
+    const swalConfig = swalLoginError(e.message);
+    if (swalConfig) {
       // in some cases is displayed loader.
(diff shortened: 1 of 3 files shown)


src/app/shared/services/login/login.service.ts
```diff
@@ -138,30 +143,34 @@ export class LoginService {
...
+   */
+  private handleLoginErrors(e, navigateOnTimeout) {
     console.error(e);
     const loginError = LOGIN_ERRORS.filter(error => e.message.includes(error.key))[0];
     if (loginError?.type === 'swal') {
-      this.displayLoginErrorWithSwal(loginError.value);
+      const config = {
+        title: 'Wrong Network',
+        text: `${loginError.value}`,
+      };
```
jrhender, line 157: It's not obvious to me why `loginError?.type === 'swal'` mentions that the error is `Wrong Network`. Does it make sense to you?

src/app/shared/services/login/login.service.ts
```diff
@@ -138,30 +143,34 @@ export class LoginService {
...
+
+      this.openSwal(config, navigateOnTimeout);
+    } else if (e.message === 'Request failed with status code 401' || e.message === 'Request failed with status code 500') {
+      const config = {
+        title: 'Session Expired',
```
jrhender, line 162: Why does status code 500 indicate "Session Expired"? That doesn't make sense to me 🤔 .

src/app/shared/services/login/login.service.ts
```diff
@@ -138,30 +143,34 @@ export class LoginService {
...
     const loginError = LOGIN_ERRORS.filter(error => e.message.includes(error.key))[0];
     if (loginError?.type === 'swal') {
-      this.displayLoginErrorWithSwal(loginError.value);
+      const config = {
+        title: 'Wrong Network',
+        text: `${loginError.value}`,
+      };
+
+      this.openSwal(config, navigateOnTimeout);
+    } else if (e.message === 'Request failed with status code 401' || e.message === 'Request failed with status code 500') {
```
jrhender, line 160: if it's possible, I think it would be better to check a `responseCode` param instead of the error message. Maybe that could be added to the error object if it isn't there already 🤔

jrhender (commented):
I like the efforts to improve the authentication experience 👍 . Some questions though that I am unsure about

src/app/shared/services/login/login.service.ts
dawidgil: Yeah, i added todo to rethink this error messaging for login. I will try today or tomorrow prepare some better approach to this.

src/app/shared/services/login/login.service.ts
dawidgil: Well, i'm not sure why is that, but when i have old token (public key in local storage) and application is trying to login again with those old values, then i get 500. Removing publicKey with wallet provider from local storage and refreshing a page is fixing this. This looks for me that backend is sending 500 when gets an invalid token.

src/app/shared/services/login/login.service.ts
jrhender: More fundamentally then the message I think, I don't really understand how `loginError?.type === 'swal'` communicates anything about the actual error. Isn't `swal` just "SweetAlert", which is just how the message is displayed? So how can we understand the substance of the error if the check is for the error display approach 😕 ? Maybe I'm missing something here though 😅

src/app/shared/services/login/login.service.ts
jrhender: Thanks for the explain. IMO, this is good place for a comment.
```suggestion
    } 
    // cache-server is returning 500 when old jwt (pubKey in local storage) is sent.
    // It seems therefore that cache-server is sending 500 when getting an invalid token.
    else if (e.message === 'Request failed with status code 401' || e.message === 'Request failed with status code 500') {
      const config = {
        title: 'Session Expired',
```

jrhender: Do you know if there is a bug logged for this in the cache-server backlog?

src/app/shared/services/login/login.service.ts
dawidgil: Yes, swal is for sweetalert package. Second one is `toastr`. My first idea was to split it into 2 ways of handling this: one with sweetalert, second with toastr. The message should be taken from `loginError.value` and which type of message from `loginError.type`. It is not the best idea imo right now. I will rewrite this :)

jrhender: Ok, do you want do that as a part of this PR, or is it important to get this PR merged in? I think it might be important to get this merged in before deploying Dev Switchboard to Volta as this login issue is quite common

jrhender: What do you think?

src/app/shared/services/login/login.service.ts
dawidgil: I don't think so, i will test this one more time + i will look if there is a bug for this. If not then i will create it.

src/app/shared/services/login/login.service.ts
dawidgil: I think i can update it in this PR. I do not know when we are going to deploy Switchboard to Volta, but this shouldn't take more than few hours of coding. Anyway in this sprint I'm mostly testing + fixing bugs that i find in SB.

src/app/shared/services/login/login.service.ts
dawidgil: Ha, not really :( This error is thrown in iam-client-lib just with a message: 
image

src/app/shared/services/login/login.service.ts
jrhender: Ok, let me know when to review the PR again then

jrhender (approved):
looks good 👍

PR #390, opened 2021-09-01
SWTCH-1189 display details of assets enrolment

dawidgil:


src/app/routes/assets/asset-enrolment-list/asset-enrolment-list.component.spec.ts
```diff
@@ -1,25 +1,114 @@
...
+import { NO_ERRORS_SCHEMA } from '@angular/core';
+import { ActivatedRoute, Params } from '@angular/router';
+import { BehaviorSubject } from 'rxjs';
+import { MockStore, provideMockStore } from '@ngrx/store/testing';
+import { UrlService } from '../../../shared/services/url-service/url.service';
+import { AssetDetailsActions, AssetDetailsSelectors } from '@state';
+import { RouterTestingModule } from '@angular/router/testing';
 
-xdescribe('AssetEnrolmentListComponent', () => {
+export class MockActivatedRoute {
```
jrhender, line 12: Is this class a generic class that could be shared by multiple tests?

src/app/shared/services/iam.service.ts
```diff
@@ -69,6 +70,10 @@ export class IamService {
     return this._iam;
   }
 
+  getAssetById(id) {
+    return this.observableWrapper(this.iam.getAssetById({id}), {message: 'Getting selected asset data...'});
```
jrhender, line 74: Is the reason for the `observableWrapper` to be able to put the `loadingService` `show()` and `hide()` around the `source`? Maybe a name like `wrapWithLoadingService` would be better?

src/app/shared/services/iam.service.ts
```diff
@@ -167,4 +172,12 @@ export class IamService {
 
     return retVal;
   }
+
+  private observableWrapper<T>(source: Promise<T> | Observable<T>,
+                               loaderConfig?: { message?: string | string[]; cancelable?: boolean }) {
+    this.loadingService.show(loaderConfig.message, !!loaderConfig.cancelable);
```
jrhender, line 178: If asserting with `!!` that `cancelable` is defined, wouldn't it better to require that in the param?
```suggestion
                               loaderConfig?: { message?: string | string[]; cancelable: boolean }) {
    this.loadingService.show(loaderConfig.message, loaderConfig.cancelable);
```

jrhender (approved):
A few comments but looks good!

src/app/routes/assets/asset-enrolment-list/asset-enrolment-list.component.spec.ts
dawidgil: Yeah, I'm planning to do a directory with mocks/stubs/function supports for tests. This require some bigger refactoring in multiple places.

src/app/shared/services/iam.service.ts
dawidgil: No, because not always it is required and not always defined.

PR #391, opened 2021-09-02
SWTCH-1171 round staking value to last 2 digits after a dot instead of taking only integer value.

dawidgil:


jrhender (approved):
I'm assuming this is just for display? Might be nice to have some tests but if it's just for display, it's probably okay

dawidgil:
After talking on standup i will ask how if should round it or take it all as it is. And answering to your question, not only displaying, this is amount of tokens that we will put to stake.

jrhender (approved):
Ok, seems good to me

PR #393, opened 2021-09-03
update iam lib and fix imports/braking changes.

dawidgil:


jrhender (changes requested):
We should wait for next version of iam-client-lib as this bug needs to be fixed: https://github.com/energywebfoundation/ew-did-registry/pull/324

We should also confirm that
- Signing in as a new user works
- Staking works
- Walletconnect works

dawidgil:
Sure i can wait with it. Just saw few breaking changes after upgrading ethers and i wanted to adjust it right now.

dawidgil:
Update will be done in different PR.

PR #395, opened 2021-09-06
SWTCH-1197 creating suborganization in suborganization view

dawidgil:
Fix a regression with not working button in sub-organization view.

### DIFF
--- a/src/app/routes/applications/governance-list/governance-list.component.html
+++ b/src/app/routes/applications/governance-list/governance-list.component.html
@@ -1,37 +1,9 @@
-<!-- Filter Sub Organization -->
-<div class="d-flex align-items-center flex-wrap mb-2" *ngIf="listType === ListType.ORG && orgHierarchy?.length">
-  <div class="d-flex align-items-center my-1" *ngIf="orgHierarchy?.length">
-    <div class="filter-list-item">
-      <a href="#" (click)="resetOrgList($event)"><span class="filter-link">ALL Organizations</span></a>
-    </div>
-    <mat-icon class="color-white opacity-50">chevron_right</mat-icon>
-  </div>
-  <ng-container *ngFor="let orgItem of orgHierarchy; let i = index;">
-    <div class="d-flex align-items-center my-1" *ngIf="i !== (orgHierarchy?.length - 1)">
-      <div class="filter-list-item">
-        <img class="img-logo-list-sm mr-2" *ngIf="!orgItem?.definition?.logoUrl"
-             src="../assets/img/no-organization-image.png"/>
-        <img class="img-logo-list-sm mr-2" *ngIf="orgItem?.definition?.logoUrl" [src]="orgItem?.definition?.logoUrl"/>
-        <a href="#" (click)="resetOrgList($event, i)"><span
-            class="filter-link">{{ orgItem?.definition?.orgName }}</span></a>
-      </div>
-      <mat-icon class="color-white opacity-50">chevron_right</mat-icon>
-    </div>
-    <div class="d-flex align-items-center my-1" *ngIf="i === (orgHierarchy?.length - 1)">
-      <img class="img-logo-list-sm mr-2" *ngIf="!orgItem?.definition?.logoUrl"
-           src="../assets/img/no-organization-image.png"/>
-      <img class="img-logo-list-sm mr-2" *ngIf="orgItem?.definition?.logoUrl" [src]="orgItem?.definition?.logoUrl"/>
-      <span class="filter-link color-white opacity-50">{{ orgItem?.definition?.orgName }}</span>
-    </div>
-  </ng-container>
-</div>
-
 <!-- START: Filter-->
 <form [formGroup]="filterForm">
-  <div *ngIf="isFilterShown && listType !== ListType.ORG"
+  <div *ngIf="isFilterShown"
        class="card card-filter d-flex justify-content-end pt-2 pb-3 mb-1">
     <div class="row">
-      <div class="col-md-3" *ngIf="listType === ListType.APP || listType === ListType.ROLE">
+      <div class="col-md-3">
         <mat-label class="color-white ml-2 pl-1">Organization</mat-label>
         <mat-form-field appearance="outline" floatPlaceholder="never"
                         class="mat-input-sort line-height-filter pt-1 mr-0 mr-lg-0 mb-2 mb-lg-0">
@@ -39,7 +11,7 @@
                  maxlength="256" (input)="filter()"/>
         </mat-form-field>
       </div>
-      <div class="col-md-3" *ngIf="listType === ListType.ROLE || listType === ListType.APP">
+      <div class="col-md-3">
         <mat-label class="color-white ml-2 pl-1">Application</mat-label>
         <mat-form-field appearance="outline" floatPlaceholder="never"
                         class="mat-input-sort line-height-filter pt-1 mr-0 mr-lg-0 mb-2 mb-lg-0">
@@ -58,8 +30,7 @@
       <div class="col-md-3" *ngIf="listType === ListType.APP"></div>
       <div class="col-md-3 d-flex justify-content-center justify-content-lg-end align-items-end mb-1">
         <button mat-stroked-button class="btn btn-small mb-0 mt-2 mt-lg-0 mb-md-2 mb-lg-0" (click)="resetFilter()">Clear
-          <ng-container *ngIf="listType !== ListType.APP">Filters</ng-container>
-          <ng-container *ngIf="listType === ListType.APP">Filter</ng-container>
+          Filters
         </button>
       </div>
     </div>
@@ -75,8 +46,6 @@
       <td mat-cell *matCellDef="let element" data-label="">
         <div class="d-flex justify-content-center justify-content-md-start">
           <ng-container *ngIf="!element?.definition?.logoUrl">
-            <img class="img-logo-list" src="../assets/img/no-organization-image.png"
-                 *ngIf="ListType.ORG === listType"/>
             <img class="img-logo-list" src="../assets/img/no-application-image.png" *ngIf="ListType.APP === listType"/>
           </ng-container>
           <img class="img-logo-list" [src]="element?.definition?.logoUrl" *ngIf="element?.definition?.logoUrl"/>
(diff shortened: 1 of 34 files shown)


src/app/routes/applications/governance-list/governance-list.component.ts
```diff
@@ -368,7 +370,27 @@ export class GovernanceListComponent implements OnInit, OnDestroy {
...
+    });
+
+    dialogRef.afterClosed()
+      .pipe(
+        filter(Boolean),
+        takeUntil(this.subscription$))
+      .subscribe(async () => await this.newSubOrg(parentOrg));
+  }
+
   async newSubOrg(parentOrg: any) {
```
jrhender, line 392: I think that `newSubOrg` and `createSubOrg` are a bit too similar a name.  How about this rename?
```suggestion
  async viewNewSubOrg(parentOrg: any) {
```

jrhender, line 392: If this is only being called in component, can it be made private?
```suggestion
  private async newSubOrg(parentOrg: any) {
```

src/app/routes/applications/governance-list/governance-list.component.ts
```diff
@@ -368,7 +370,27 @@ export class GovernanceListComponent implements OnInit, OnDestroy {
...
+      },
+      maxWidth: '100%',
+      disableClose: true
+    });
+
+    dialogRef.afterClosed()
+      .pipe(
+        filter(Boolean),
+        takeUntil(this.subscription$))
+      .subscribe(async () => await this.newSubOrg(parentOrg));
```
jrhender, line 389: This is probably due to my lack of RxJs knowledge but could you help me understand what this code is doing?

- In `pipe` you only take the not defined/not-null values (due to the `filter(Boolean)`). What happens if the return from `afterClosed()` is `undefined`?
- In `pipe`, `takeUntil(this.subscription$)` is a little unclear. Like I get that it means take until `this.subscription$` emits, but why should we do this? What is `this.subscription$` observing? Is this an Angular convention?
- `this.newSubOrg` is called on the `subscribe`. Does this mean that in theory `this.newSubOrg` could be called more than once?

src/app/routes/applications/governance-list/governance-list.component.ts
```diff
@@ -368,7 +370,27 @@ export class GovernanceListComponent implements OnInit, OnDestroy {
...
+        parentOrg: JSON.parse(JSON.stringify(parentOrg)),
+        owner: parentOrg.owner
+      },
+      maxWidth: '100%',
+      disableClose: true
+    });
+
+    dialogRef.afterClosed()
+      .pipe(
+        filter(Boolean),
```
jrhender, line 387: I was reading a bit about `filter(Boolean)` and found this SO answer which suggests using a explicit filter for `undefined` so that the type is preserved. Have you seen this before? https://stackoverflow.com/a/57928681
```suggestion
        filter(v=> v!== undefined),
```

jrhender (commented):
A few questions/suggestions. Probably I'm a bit to unfamiliar with Angular/RxJs 😓

src/app/routes/applications/governance-list/governance-list.component.ts
dawidgil: Ok, sure i can rename it. I just wanted to remove regression before deploying to volta and handle it in next few days properly.

src/app/routes/applications/governance-list/governance-list.component.ts
jrhender: Ah, maybe it's better to just provide the type like `filter<type>(Boolean)` but maybe that's not necessary in this case?

dawidgil: hmm, interesting. Thanks for sharing this one.

src/app/routes/applications/governance-list/governance-list.component.ts
dawidgil: This is used in template as well.
I refactored this once and that's why this regression happened. I didn't expected that parent component is using child component methods.
I created there components for actions, and those components are handling displaying dialogs etc. method like `newSubOrg` is used in template just to refresh the list after closing a dialog.

src/app/routes/applications/governance-list/governance-list.component.ts
dawidgil: Ah now i remember. Filtering is needed because we can send there some values while closing dialogs. We are sending true/false values there. True when we want to refresh data and false if just closing a dialog.

src/app/routes/applications/governance-list/governance-list.component.ts
jrhender: Ok, fair enough 👍

src/app/routes/applications/governance-list/governance-list.component.ts
jrhender: Ok, thanks for the explanations

dawidgil:
@jrhender I refactored a lot here and fixed few bugs/inconsistent behavior. I used ngrx to handle most of things in organization list view. I also created a new component for handling only organization list. I want to do the same with applications and roles in future and extract common stuff to abstract class. I also added tests to most of things.

src/app/routes/applications/organization-list/organization-list.component.html
```diff
@@ -0,0 +1,114 @@
+<!-- Filter Sub Organization -->
```
jrhender, line 1: Does it make sense to you that this new `organization-list.component` is under `routes/applications`. What does "applications" mean in this context 🤔 ?

src/app/state/organization/utils/get-main-orgs.ts
```diff
@@ -0,0 +1,40 @@
...
+      }
+      subList[namespaceArr.length - 4].push(item);
+    }
+  });
+
+  // Remove Unnecessary Sub-Orgs from Main List
+  if (list.length || subList.length) {
+    for (let i = 0; i < subList.length; i++) {
+      const arr = subList[i];
+      if (arr && arr.length) {
```
jrhender, line 22: The subsequent code doesn't seem to be referencing the `length` property of `arr`? Why the check for it in this `if`?

src/app/state/organization/utils/get-main-orgs.ts
```diff
@@ -0,0 +1,40 @@
...
+    const namespaceArr = item.namespace.split('.');
+    if (namespaceArr.length === 3) {
+      list.push(item);
+    } else {
+      if (!subList[namespaceArr.length - 4]) {
+        subList[namespaceArr.length - 4] = [];
+      }
+      subList[namespaceArr.length - 4].push(item);
+    }
+  });
```
jrhender, line 16: The `3` and `4` are a bit of a magic numbers. I think constants should be used.
I'm not sure why the `4`. If you had a suborg of `finance.dawidcorp.iam.ewc`, `namespaceArr.length` would be `4`. So I guess what you're trying to do is get lists of the suborg that are at each "level"?
```suggestion
  // Separate Parent & Child Orgs
   const numberOfLabelsInOrgNamespace = 3;
  $getOrgList.forEach((item: any) => {
    const namespaceArr = item.namespace.split('.');
    if (namespaceArr.length === numberOfLabelsInOrgNamespace) {
      list.push(item);
    } else {
      if (!subList[namespaceArr.length - 4]) {
        subList[namespaceArr.length - 4] = [];
      }
      subList[namespaceArr.length - 4].push(item);
    }
  });
```

src/app/state/organization/utils/get-main-orgs.ts
```diff
@@ -0,0 +1,40 @@
+export const getMainOrgs = ($getOrgList: any[]) => {
```
jrhender, line 1: I find this function pretty confusing, maybe it could use some documentation

src/app/state/organization/utils/get-main-orgs.ts
```diff
@@ -0,0 +1,40 @@
...
+          let exists = false;
+          for (const mainOrg of list) {
+            if (subOrg.namespace && subOrg.namespace.includes(mainOrg.namespace)) {
+              exists = true;
+              break;
+            }
+          }
+          if (!exists) {
+            list.push(subOrg);
+          }
```
jrhender, line 33: It seems like you are trying to figure out if a `subOrg`'s parent Org is in the `list`? Is it possible to have different result based on the order of the elements in `arr`? Is you're checking the elements of `list` but also adding to `list` in each iteration?

jrhender (commented):
Looks good, I think the `get-main-orgs` function could just use some additional clarification

src/app/routes/applications/governance-list/governance-list.component.html
```diff
@@ -138,18 +97,6 @@
...
-                                      (organizationCreated)="newSubOrg(element)"
-                                      (viewApps)="viewApps(ListType.ORG, element)"
-                                      (viewRoles)="viewRoles(ListType.ORG, element)"
-                                      (appCreated)="newApp(element)"
-                                      (roleCreated)="viewRoles(ListType.ORG, element)"
-                                      (edited)="edit()"
-                                      (transferred)="transferOwnership()"
-                                      (deleteConfirmed)="remove(listType, element)">
-            </app-organization-actions>
-
```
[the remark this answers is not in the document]
dawidgil: No governance-list is for application and roles right now. organization-list is only for organization. I didn't want to refactor everything, but the plan is to split application and roles into separate components.

src/app/routes/applications/organization-list/organization-list.component.html
dawidgil: No, it doesn't. I forgot to change application.component.ts into governance.component.ts. I should change also routing here. Thanks for pointing this out!

src/app/state/organization/utils/get-main-orgs.ts
dawidgil: Ok, this code is just extracted from governance-list.component.ts :D I didn't touched that. I just wrote few tests to be sure that it is working correctly. I will refactor this.
To let you know what is going on here, this code is getting main organization (for example test.iam.ewc) and all it's direct suborganization (suborg.test.iam.ewc etc.), but not sub-(...)-sub-org.

jrhender (approved):
👍

PR #401, opened 2021-09-15
SWTCH-1190 transfer asset ownership should use address.

dawidgil:
This PR solves also bugs with Ethereum address validation.


jrhender (approved):
Looks good to me, thanks!

PR #403, opened 2021-09-17
fix(auth): SWTCH-1205 race condition when initializing staking pool.

dawidgil:
The error occurs when you have some organizations and tries to go directly to [switchboard-dev.energyweb.org/governance](switchboard-dev.energyweb.org/governance) page. There is a race condition, which will be first: initialization of staking pool or getting list of organizations which cause an error. This PR solves problem by redirecting to governance page after successful staking pool initialization.

### DIFF
--- a/src/app/state/stake/stake.effects.spec.ts
+++ b/src/app/state/stake/stake.effects.spec.ts
@@ -15,7 +15,7 @@ import * as PoolActions from '../pool/pool.actions';
 import { skip, take } from 'rxjs/operators';
 import { StakingPoolServiceFacade } from '../../shared/services/staking/staking-pool-service-facade';
 import { StakingPoolFacade } from '../../shared/services/pool/staking-pool-facade';
-import { LayoutActions } from '@state';
+import * as LayoutActions from '../layout/layout.actions';
 
 describe('StakeEffects', () => {
 
--- a/src/app/state/stake/stake.effects.spec.ts
+++ b/src/app/state/stake/stake.effects.spec.ts
@@ -54,7 +54,7 @@ describe('StakeEffects', () => {
       actions$ = new ReplaySubject(1);
     });
 
-    it('should return initPool action and getAccountBalance action', waitForAsync(() => {
+    it('should return initPool action, getAccountBalance action and redirect action', waitForAsync(() => {
       actions$.next(StakeActions.initStakingPool());
       stakingService.init.and.returnValue(of(true));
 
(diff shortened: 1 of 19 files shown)


src/app/state/stake/stake.effects.spec.ts
```diff
@@ -63,9 +62,13 @@ describe('StakeEffects', () => {
...
       });
 
-      effects.initStakingPoolService$.pipe(skip(1)).subscribe(resultAction => {
+      effects.initStakingPoolService$.pipe(skip(1), take(1)).subscribe(resultAction => {
         expect(resultAction).toEqual(PoolActions.getAccountBalance());
       });
+
+      effects.initStakingPoolService$.pipe(skip(2), take(1)).subscribe(resultAction => {
+        expect(resultAction).toEqual(LayoutActions.redirect());
+      });
```
jrhender, line 71: I think it could be good to have a comment here as what `skip(2)` vs `skip(1)` represent from a functional standpoint. Why does `skip(1)` mean that you could `getAccountBalance` and `skip(2)` mean that you should `redirect`? Maybe saying that this behaviour "is to fix a race condition between initialization of staking pool and getting list of organizations" would be helpful? I'm not sure if this suggestion below is correct but maybe it gives a starting point?
```suggestion
      /**
       * First we get account balance and then redirect in order to to fix a race condition between initialization of staking pool and getting list of organizations
       */
      effects.initStakingPoolService$.pipe(skip(1), take(1)).subscribe(resultAction => {
        expect(resultAction).toEqual(PoolActions.getAccountBalance());
      });
      effects.initStakingPoolService$.pipe(skip(2), take(1)).subscribe(resultAction => {
        expect(resultAction).toEqual(LayoutActions.redirect());
      });
```

dawidgil: account balance is not relevant for the fix. it is needed for staking, but it is called after successful initialization of staking pool. Fix for condition race lays in `redirect` action. So I moved redirect to be called after initialization staking pool instead of successful login like it was previously (race between redirect and initialization staking pool).

Skip part is for skipping previous actions. For example if i want to check third dispatched action, i need to skip two of them. I need to dive into marble testing, because it is easier to read/understand but I couldn't make it work with effects (it's working when i create a custom operator).

However you are right, i will add a comment but in effect file why that redirect is needed.

FYI Dispatching actions is like calling an event (clicking), it's asynchronous and we do not know when it fulfills.

jrhender (approved):
Looks good to me!

PR #406, opened 2021-09-20
SWTCH-1207 connect role summary design with data

dawidgil:
Updates role creation fields summary (confirm details step) with new design. Reuses the same component in two places.


davycastillo (approved):
Lgtm.

PR #407, opened 2021-09-20
SWTCH-1206 tests utils folder

dawidgil:


tsconfig.spec.json
```diff
@@ -12,6 +12,9 @@
     "src/test.ts",
     "src/polyfills.ts"
   ],
+  "exclude": [
+    "src/app/tests/**.ts"
+  ],
```
jrhender, line 17: This change doesn't make alot of sense to me. I'm assuming the `tsconfig.spec.json` is the tsconfig that is used for tests, so `src/app/tests/**.ts` would be a part of that?

tsconfig.json
```diff
@@ -27,6 +27,9 @@
     "paths": {
       "@state": [
         "src/app/state/index"
+      ],
+      "@tests": [
+        "src/app/tests/index"
```
jrhender, line 32: This also doesn't much sense to me. Is it because `@tests` is being imported from non test files?

jrhender (commented):
Seems like some good consolidation to me, just a couple questions

tsconfig.json
dawidgil: This is alias to file. In this case to index.ts file which exports all mocks/stubs etc. That way, i can remove paths like `../../../../../../something/.../file.ts`

tsconfig.spec.json
dawidgil: Maybe i'm wrong, but my intention was to exclude this part of code from code coverage. But after checking this, you are right, this make no difference. So i will remove this part.

tsconfig.json
jrhender: Ah yes, right, I think you might have explained this to me before 😆

jrhender (approved):

PR #410, opened 2021-09-21
chg: move custom rxjs operators to one folder

dawidgil:


jrhender:
In general, I prefer grouping files by feature or domain area, not by type https://mnapoli.fr/organizing-code-into-domain-modules/
This seems to be move the code in a "group by type" direction? It seems like some of the operators are generic and don't really belong to a feature area (e.g. `truthy` operator) but some do seem to be related to a feature area (e.g. `map-claim-profile`, `filter-provider`). So maybe this isn't a good change 🤔 ? What do you think @dawidgil ?

dawidgil:
In fact mapClaimsProfile operator is used in assets, and in user claim state slice. So it is used in two independent places.
Probably moving `filter-provider` file is not a good idea, so i revert this one.

jrhender (approved):
Ok, sounds good!

PR #412, opened 2021-09-21
fix: SWTCH-1209 opening dialog for suborganization that do not belong to us

dawidgil:
Button for creation sub organization was enabled always, even if we were looking on organization which namespace belongs to a different user.


jrhender (approved):
Nice fix 👍

PR #420, opened 2021-09-28
SWTCH-1210 split governance view into multiple components

dawidgil:
Splits governance view into separate components. Now we have organization-list, application-list, role-list components. Each is responsible for its own view. Added tests.


davycastillo (approved):
Lgtm.

PR #425, opened 2021-10-04
SWTCH-1211 initialize iam client lib with private key in e2e tests

dawidgil:
To run you need to specify `privateKey` in `environment.ts` file and run script `e2e:open`. It will open a cypress electron app where you can run tests in browser. privateKey property can be the same as in metamask (just export it from MM). 
This first approach mocks backend for owned asset list.

I want to achieve at least what is below. If there are any other ideas what should i do please, let me know.

- [ ] Basic tests for Assets for displaying / validation with mocked backend.
- [ ] if possible: test with mocked backend interaction with blockchain, for example Registering Asset. 
- [ ] Configure CI with GHA.


dawidgil:
I will close for now this PR, to not spam your emails with testing CI in GHA.

PR #426, opened 2021-10-05
chg(deps): Update angular and related packages.

dawidgil:
Updates angular to version 12 where is used webpack 5. Webpack 5 do not include node polyfills anymore, so there was a need to inject them.


dawidgil:
Closing as i will create new PR with Angular update.

PR #427, opened 2021-10-06
feature/login with EKC

JGiter:


dawidgil:
Adding async/await won't help, because rxjs is handling asynchronous calls already. So there is no need to change that.
I'm not able to reproduce this, because i have problems with login via Azure button. 🤔 
However i saw in method `closeConnection` in `iam-base.ts` was removed a line of code `await this._walletConnectService.closeConnection();`. Maybe because of that you can't logout?
Anyway i don't think it's a frontend problem, because we just simply using iam-client-lib API.

dawidgil (approved):

PR #435, opened 2021-10-12
SWTCH-782 generate qrcode for assets

dawidgil:


jrhender:
@dawidgil What is the expected behaviour here for testing? If I scan with a QR code app on a phone the asset DID should come up?

dawidgil:
Yes, when you scan this generated qr code, then you should get asset did.
We do not have scanning yet in application, so you can't test it with switchboard. However if you use some other tool, this should work.

PR #458, opened 2021-11-05
SWTCH-1215 issuance of role vcs

dawidgil:


src/app/routes/registration/utils/precondition-check.ts
```diff
@@ -0,0 +1,63 @@
...
+      if (roleObj.isAccepted) {
+        if (roleObj.isSynced) {
+          status = RolePreconditionType.SYNCED;
+        } else {
+          status = RolePreconditionType.APPROVED;
+        }
+      }
+      break;
+    }
+  }
```
jrhender, line 60: We also need a check here about `on-chain` vs `off-chain`. Because an on-chain role would be not be "synced" but it doesn't need to be synced.

Actually, testing locally I ran into this issue. I requested (and was issued) an on-chain role only and SB told me to sync it, which shouldn't be the case.
image

jrhender (changes requested):
It looks good, just a potential issue with the "isSynced" logic not taking into account whether or not the role is off-chain vs on-chain

src/app/routes/registration/utils/precondition-check.ts
dawidgil: Ok, so the problem is with displaying message about `Pending Sync to DID Document` right? It should be not possible to sync if there is an on-chain option selected. What if there is on-chain and off-chain selected? The same case -> no need to sync?
I'm not sure if the problem lays in this selected part of code, because it have nothing to do with displaying status of an enrollment.

jrhender: Hmm, yes, the issue might not be in this code. I was thinking that it might be because a role that is `on-chain` only might not have `isSynced==true` but still wouldn't need to be synced. But the issue could definitely be elsewhere.

As far as the logic, an `on-chain` credential does not need to be synced, while an 'off-chain` credential does need to be synced. So:
- If user has on-chain credential: nothing needs to be synced
- If user has off-chain credential: off-chain credential needs to be synced
- If user has both on-chain and off-chain: only off-chain needs to be synced

src/app/modules/issue-vc/new-issue-vc/new-issue-vc.component.html
```diff
@@ -6,65 +6,49 @@ <h4 mat-dialog-title class="mb-0 d-flex justify-content-between align-items-cent
...
-              <mat-option *ngFor="let role of roles"
-                          [value]="role" [attr.data-qa-id]="role.name">{{role.name}}</mat-option>
-            </mat-select>
-          </mat-form-field>
+          <div class="card card-default d-flex shadow-none pt-3 mb-3">
+            <mat-label class="ml-3 pl-1">Subject</mat-label>
+            <mat-form-field class="mt-2" appearance="outline">
+              <input matInput formControlName="subject" placeholder="DID that is receiving the credential"
+                     maxlength="256" data-qa-id="subject-did" required autocomplete="off">
+              <button *ngIf="!isDidPredefined()" matSuffix mat-icon-button aria-label="Scan DID" appQrCodeScanner
```
Malak67, line 18: Usually is not a good approach to call method in template. I would use a getter in the .ts file and use that here:
```
get isDidPredefined() {
}

ngIf="isDidPredefined"
```

src/app/modules/issue-vc/new-issue-vc/new-issue-vc.component.html
```diff
@@ -6,65 +6,49 @@ <h4 mat-dialog-title class="mb-0 d-flex justify-content-between align-items-cent
...
-                   required autocomplete="off">
-            <button *ngIf="!isDidPredefined()" matSuffix mat-icon-button aria-label="Scan DID" appQrCodeScanner
-                    (scannedValue)="scannedValue($event)">
-              <mat-icon>qr_code_2</mat-icon>
-            </button>
-            <mat-error *ngIf="getControl('subject').hasError('required')">Subject is <strong>required</strong>
-            </mat-error>
-            <mat-error *ngIf="getControl('subject').hasError('invalidDid')">Subject DID format is invalid</mat-error>
-          </mat-form-field>
+        <div class="col-lg-12" *ngIf="isFormSubjectValid()">
```
Malak67, line 28: Same comment as before. I would use a getter function

src/app/modules/issue-vc/new-issue-vc/new-issue-vc.component.ts
```diff
@@ -21,175 +19,82 @@ export class NewIssueVcComponent implements OnInit {
...
-      if (this.form.get('subject').valid) {
-        await this.getNotEnrolledRoles(d);
-      }
-
-    });
+    this.setPossibleRoles();
   }
 
-  async roleTypeSelected(e: any) {
+  roleTypeSelected(e: any) {
```
Malak67, line 38: I would try to avoid using "any".  You should be able to see the type or just console.log(e) and see it there.

src/app/modules/issue-vc/new-issue-vc/new-issue-vc.component.ts
```diff
@@ -21,175 +19,82 @@ export class NewIssueVcComponent implements OnInit {
...
     if (e && e.value && e.value.definition) {
```
Malak67, line 39: I think this would look better:
`if (e?.value?.definition)`

src/app/modules/issue-vc/new-issue-vc/new-issue-vc.component.ts
```diff
@@ -21,175 +19,82 @@ export class NewIssueVcComponent implements OnInit {
...
 
-  isFormDisabled() {
-    return this.form.invalid || this.alreadyEnroled || !this.isPrecheckSuccess;
+  isFormDisabled(): boolean {
+    return this.form.invalid || !this.isPrecheckSuccess;
   }
 
-  create() {
-    if (this.isFormDisabled()) {
+  create(e) {
```
Malak67, line 74: I would add a type here as well:
`e: Event-blabla`

src/app/modules/issue-vc/new-issue-vc/new-issue-vc.component.ts
```diff
@@ -21,175 +19,82 @@ export class NewIssueVcComponent implements OnInit {
...
+    this.issuanceVcService.create({subject: this.getFormSubject().value, claim: this.createClaim(e.fields)})
+      .subscribe(() => this.dialogRef.close());
+  }
+
+  private setPossibleRoles() {
+    this.getFormSubject().valueChanges
+      .pipe(
+        filter(() => this.isFormSubjectValid()),
+        switchMap((did) => this.issuanceVcService.getNotEnrolledRoles(did)),
+      ).subscribe((list) => this.possibleRolesToEnrol = list);
```
Malak67, line 87: Type missing

src/app/modules/issue-vc/new-issue-vc/new-issue-vc.component.ts
```diff
@@ -21,175 +19,82 @@ export class NewIssueVcComponent implements OnInit {
...
-      });
+    this.issuanceVcService.create({subject: this.getFormSubject().value, claim: this.createClaim(e.fields)})
+      .subscribe(() => this.dialogRef.close());
+  }
+
+  private setPossibleRoles() {
+    this.getFormSubject().valueChanges
+      .pipe(
+        filter(() => this.isFormSubjectValid()),
+        switchMap((did) => this.issuanceVcService.getNotEnrolledRoles(did)),
```
Malak67, line 86: Type missing

src/app/modules/issue-vc/new-issue-vc/new-issue-vc.component.ts
```diff
@@ -21,175 +19,82 @@ export class NewIssueVcComponent implements OnInit {
...
   private setDid() {
     if (this.data?.did) {
       this.form.patchValue({subject: this.data.did});
-      this.form.get('subject').disable();
+      this.getFormSubject().disable();
     }
   }
 
-  private createClaim() {
+  private createClaim(fields) {
```
Malak67, line 97: Type missing

src/app/modules/issue-vc/new-issue-vc/new-issue-vc.component.ts
```diff
@@ -198,9 +103,9 @@ export class NewIssueVcComponent implements OnInit {
     };
 
     return {
-      fields: this.fieldList,
-      claimType: this.selectedNamespace,
-      claimTypeVersion: parseVersion(this.selectedRole.version) || DEFAULT_CLAIM_TYPE_VERSION
+      claimType: this.getFormType().value.namespace,
+      claimTypeVersion: parseVersion(this.selectedRoleDefinition.version) || DEFAULT_CLAIM_TYPE_VERSION,
```
Malak67, line 107: Is this working? I am thinking if this.selectedRoleDefinition.version is missing, then parseVersion might return error. Not sure. Might be good to do a ternary operator here:
`this.selectedRoleDefinition.version ? parseVersion(this.selectedRoleDefinition.version) : DEFAULT_CLAIM_TYPE_VERSION,`

src/app/modules/required-fields/components/required-fields/required-fields.component.html
```diff
@@ -0,0 +1,19 @@
+<form [formGroup]="form">
+  <ng-container *ngFor="let field of getFields()?.controls; let i = index;">
```
Malak67, line 2: A getter would work better here as well.

src/app/routes/applications/new-role/components/fields.abstract.ts
```diff
@@ -0,0 +1,80 @@
...
+export abstract class FieldsAbstract<T> {
+  abstract updateData: EventEmitter<T>;
+
+  abstract get fields(): T[];
+
+  showFieldsForm = false;
+  fieldIndex: number;
+  fieldToEdit = null;
+
+  addFieldHandler(fieldData) {
```
Malak67, line 12: a type would be nice here.

src/app/routes/enrolment/view-requests/view-requests.component.html
```diff
@@ -45,32 +45,22 @@ <h3 class="color-white mb-2">{{ claim?.roleName | uppercase }}</h3>
...
+        <app-fields *ngIf="issuerFields?.length" title="Issuer fields" [fieldsList]="issuerFields"></app-fields>
+
         <div class="col-lg-12 mt-2" *ngIf="!fields?.length">
           <div class="card card-info d-flex flex-row justify-content-start align-items-center shadow-none d-flex mb-3">
             <mat-icon>info</mat-icon>
             <span class="f-14">Role did not require any field for enrolment request.</span>
           </div>
         </div>
-        <div class="col-lg-12 mt-2 mb-3" *ngIf="canAccept()">
+        <div class="col-lg-12 mt-2 mb-3" *ngIf="canAccept() && roleContainRequiredParams()">
```
Malak67, line 57: I would use getters or just norml props insted of calling methods here.

src/app/routes/enrolment/view-requests/view-requests.component.spec.ts
```diff
@@ -11,7 +11,7 @@ import { NotificationService } from '../../../shared/services/notification.servi
 import { dialogSpy, iamServiceSpy, loadingServiceSpy } from '@tests';
 import { of } from 'rxjs';
 
-describe('ViewRequestsComponent', () => {
+xdescribe('ViewRequestsComponent', () => {
```
Malak67, line 14: is this intended or you don`t whish to run this test?

Malak67 (commented):
Left some comments.

src/app/modules/issue-vc/new-issue-vc/new-issue-vc.component.html
dawidgil: Why it is not a good approach? I know that it is calculated on each change detection cycle, but both solutions are. There are not benefits from this. It's more like a syntactic sugar here :D

src/app/routes/enrolment/view-requests/view-requests.component.spec.ts
dawidgil: I disabled this test because it was broken. I adjusted code but didn't add tests here. For me this component should be split into smaller parts and those should be tested. I spent on this task a lot of time and didn't want to refactor/adjust everything :) This is one of the compromises that i made. However i can fix running tests for this component.

Malak67: Ok, makes sense.

src/app/modules/issue-vc/new-issue-vc/new-issue-vc.component.html
Malak67: Yeah, pipes or variables would be better. Calling methods would end up in performance issues due to change detection. So I would try to use pipes and variables where possibile. I prefer getters to calling methods mostly from a visual perspective.

src/app/modules/issue-vc/new-issue-vc/new-issue-vc.component.ts
dawidgil: As far as i know, version always exist. This is kinda copy paste from different place [Look here](https://github.com/energywebfoundation/switchboard-dapp/blob/382bd85e2a3e2d0c347db0648636cfd6f71196ee/src/app/routes/registration/request-claim/request-claim.component.ts#L316-L332). This is another compromise - to not change/adjust different part of code where there is no tests. I will add TODO to change this in the future :)

jrhender (approved):
I'm not able to fully test due to a dev cache-server issue, but the code looks fine to me

PR #459, opened 2021-11-09
SWTCH-1236 did phonebook

dawidgil:


src/app/utils/functions/extract-address/extract-address.ts
```diff
@@ -0,0 +1,8 @@
+import { didRegex } from '../../validators/is-hex/is-hex.validator';
+
+export const extractAddress = (did: string): string => {
+  if (didRegex.test(did.trim())) {
+    return did.split(':')[2];
```
jrhender, line 5: This would be more future-proof, given the changes that we are making here https://energyweb.atlassian.net/browse/EDR-88
```suggestion
    return did.split(':').pop();
```

jrhender (changes requested):
A minor point to future-proof the solution a bit

src/app/utils/functions/extract-address/extract-address.ts
jrhender: I think this makes sense because I think, assuming there is no fragment, then the last section will be the `method-specific-id` https://www.w3.org/TR/did-core/#did-syntax
Does this make sense @nichonien ?

dawidgil: Yeah, you're right. It is better solution, because we always wants last part. I will adjust it. Thanks!

jrhender (approved):
Looks great, thanks!

PR #461, opened 2021-11-16
fix: notification after rejecting signature in MM.

dawidgil:


src/app/routes/enrolment/view-requests/view-requests.component.ts
```diff
@@ -80,7 +80,7 @@ export class ViewRequestsComponent implements OnInit {
       this.toastr.success('Request is approved.', TOASTR_HEADER);
       this.dialogRef.close(true);
     } catch (e) {
-      this.toastr.error(e, TOASTR_HEADER);
+      this.toastr.error(e.message, TOASTR_HEADER);
```
Harasz, line 83:
```suggestion
      this.toastr.error(e?.message, TOASTR_HEADER);
```

Harasz: I think it would be safer to check for a message param

Harasz (approved):

PR #463, opened 2021-11-17
ci: cloudfront deployments

hejkerooo:


.github/workflows/cf-deploy.yml
```diff
@@ -0,0 +1,77 @@
...
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.AWS_REGION }}
+      - name: Clear bucket
+        run: |
+          aws s3 rm s3://${{ secrets.BUCKET_NAME }} --recursive
+      - name: Upload to bucket
+        run: |
+          aws s3 cp dist s3://${{ secrets.BUCKET_NAME }} --recursive
+      - name: Invalidate CloudFront
```
jrhender, line 75: Why this `Invalidate Cloudfront` step?

jrhender (commented):
Thanks for the PR @hejkerooo . Are you going to be taking on https://energyweb.atlassian.net/browse/DOI-1439 or is this just a PR to share your code?

.github/workflows/cf-deploy.yml
hejkerooo: Yes, I'm taking care of it.

Cloudfront caches s3 content so after each deployment, it needs to be cleared.

jrhender: Thanks, what do you think of adding this as a comment?
```suggestion
      # cloudfront caches s3 content so after each deployment, it needs to be cleared.
      - name: Invalidate CloudFront
```

.github/workflows/cf-deploy.yml
```diff
@@ -0,0 +1,79 @@
...
+        run: |
+          aws cloudfront create-invalidation --distribution-id ${{ secrets.CLOUDFRONT_ID }} --paths "/*"
```
jrhender, line 79: Will each environment of SB have its own `distribution-id`/`CLOUDFRONT_ID`, so a deploy to one environment won't invalidate the cache of the other environments?

jrhender (commented):
Just a couple more comments

.github/workflows/cf-deploy.yml
hejkerooo: Yes

dawidgil (approved):

PR #470, opened 2021-11-23
SWTCH-1247 specify format

dawidgil:


src/app/routes/registration/enrolment-form/enrolment-form.component.spec.ts
```diff
@@ -119,6 +119,62 @@ describe('EnrolmentFormComponent', () => {
...
+
+      const {offChain, onChain} = getSelectors(hostDebug);
+
+      expect(offChain.nativeElement.checked).toBeTruthy('off chain should be selected');
+      expect(offChain.nativeElement.disabled).toBeTruthy('off chain is not disabled');
+      expect(onChain.nativeElement.checked).toBeFalsy('on chain should be deselected');
+      expect(onChain.nativeElement.disabled).toBeTruthy('on chain is not disabled');
+    });
+
+    it('should disable checkboxes when offChain and onChain are predefined and set to true', () => {
```
jrhender, line 151: Should this say `enable`?
```suggestion
    it('should enable checkboxes when offChain and onChain are predefined and set to true', () => {
```

src/app/routes/registration/enrolment-form/enrolment-form.component.spec.ts
```diff
@@ -119,6 +119,62 @@ describe('EnrolmentFormComponent', () => {
...
+      component.fieldList = [];
+      component.predefinedRegTypes = {offChain: true, onChain: true};
+
+      fixture.detectChanges();
+
+      const {offChain, onChain} = getSelectors(hostDebug);
+
+      expect(offChain.nativeElement.checked).toBeTruthy('off chain should be selected');
+      expect(offChain.nativeElement.disabled).toBeTruthy('off chain is disabled');
```
jrhender, line 160: These messages are the error messages, right?
```suggestion
      expect(offChain.nativeElement.disabled).toBeTruthy('off chain is not disabled');
```

jrhender (commented):
Just some comments about the messages

src/app/routes/registration/enrolment-form/enrolment-form.component.spec.ts
dawidgil: No, what i understood checkboxes should be disabled when are predefined by url. So if we have in url onchain=true then checkbox for on-chain should be disabled but checked. Checkbox for off-chain should be also disabled but unchecked (disabled means when i click on it, nothing happens)

src/app/routes/registration/enrolment-form/enrolment-form.component.spec.ts
dawidgil: yes, useful for debugging.

src/app/routes/registration/enrolment-form/enrolment-form.component.spec.ts
jrhender: Oh yes, you are right 👍

dawidgil:
Adjusted error information in unit tests. I used one approach which is `checkbox should be/not be disabled` to remove confusion. Hope that now it looks good :)

jrhender (approved):
Looks great, thanks!

PR #471, opened 2021-11-23
SWTCH-1234 dsla tasks

dawidgil:
This PR provides few changes:
- limit to 50,000 EWT per DID - this information is from iam-client-lib
- validation of stake value - do not allow to put stake above maximum value
- allow to stake multiple times
- information about taken gas price for transaction

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1229,9 +1229,9 @@
       }
     },
     "@babel/helper-define-polyfill-provider": {
-      "version": "0.2.4",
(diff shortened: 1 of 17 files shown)


jrhender (approved):
I'm not sure if everything is correct as I'm not as familiar with this story.
I have one question about the empty `of()` but I'm not that familiar with RxJs, so maybe it's fine 👍 
Feel free to merge from my perspective

dawidgil:
`of` is because they changed staking pool contracts. At this moment there is no launchPool method in StakingFactoryService interface. Maybe it will be in future, but not now. Didn't want to remove everything, because it might be needed at some point.

PR #474, opened 2021-11-23
SWTCH-1254 add: optional method specific id to regex.

dawidgil:


src/app/utils/validators/is-hex/is-hex.validator.spec.ts
```diff
@@ -104,6 +104,10 @@ describe('tests for isHexValidator', () => {
     it('should return null when did is valid', () => {
       expect(getFormErrors('did:ethr:0x' + stringWithLength(40), HexValidators.isDidValid)).toEqual(null);
     });
+
+    it('should return null when did have method and method specified id', () => {
+      expect(getFormErrors('did:ethr:specificid:0x' + stringWithLength(40), HexValidators.isDidValid)).toEqual(null);
```
jrhender, line 109:
```suggestion
      expect(getFormErrors('did:ethr:chainid:0x' + stringWithLength(40), HexValidators.isDidValid)).toEqual(null);
```

jrhender (approved):
Small comment but looks good

PR #478, opened 2021-11-25
feat: handle partial withdraw. Add info about staking pool expiration.

dawidgil:


src/env.js
```diff
@@ -13,6 +13,6 @@
     window.__env.cacheServerUrl = 'https://identitycache-dev.energyweb.org/v1';
     window.__env.natsServerUrl = 'https://identityevents-dev.energyweb.org/';
     window.__env.kmsServerUrl = undefined;
-    window.__env.stakingPoolFactoryAddress = '0x3490B3e5E7C1f696c61FDe1c2679725B19274B0e';
+    window.__env.stakingPoolFactoryAddress = '0xd61Ac65c0fECfb83108f6E374a57FfF9c692901d';
```
Malak67, line 16: Shouldn`t this stuff be taken from a .env file and not commited? Only if it`s sensitive data. If not, they maybe it`s ok

src/app/routes/ewt-patron/withdraw/withdraw.component.html
```diff
@@ -1,18 +1,35 @@
...
-    </div>
+<div class="d-flex justify-content-center align-items-center flex-column stake-success-container">
+  <h3 class="mt-3 mb-4">Withdraw?</h3>
+  <p class="text-center mt-1">
+    EWT will be transfered to your account balance.
+    You will be charged with gass fee.
+  </p>
+</div>
+
+<mat-form-field class="staking-textfield" appearance="outline">
```
Malak67, line 9: Usually it is a good approach to keep all form fields inside a form, even if it's only one.

src/app/shared/services/pool/staking-pool-facade.ts
```diff
@@ -21,11 +21,21 @@ export class StakingPoolFacade {
...
+  getStartDate() {
+    return from(this.pool.getStart());
+  }
+
+  getEndDate() {
+    return from(this.pool.getEnd());
+  }
+
   checkReward() {
+    console.log('check reward');
```
Malak67, line 33: Is this intentional or missed here? Better to remove console.logs before merging but maybe you need to check something after it is deployed?

src/app/shared/services/pool/staking-pool-facade.ts
```diff
@@ -21,11 +21,21 @@ export class StakingPoolFacade {
...
     return from(this.pool.checkReward());
   }
 
   getStake() {
+    console.log('get stake');
```
Malak67, line 38: Same question as above.

src/app/state/pool/pool.selectors.ts
```diff
@@ -10,7 +10,7 @@ export const getStakeState = createFeatureSelector<PoolState>(USER_FEATURE_KEY);
 
 export const getReward = createSelector(
   getStakeState,
-  (state: PoolState) => state.reward
+  (state: PoolState) => formatEther(state.reward as any)
```
Malak67, line 13: Why cast it as any? Can't you create a type or interface for it?

src/env.js
dawidgil: It should, but right now devops do not take care of it. So it is maintained by repo.

src/app/shared/services/pool/staking-pool-facade.ts
dawidgil: missed, thanks for catching this!

Malak67 (approved):

PR #484, opened 2021-11-30
fix: unhandled error while trying to initialize iam.

dawidgil:
The error occurs while trying to initialize iam, but error was thrown. Iam service is catching that error while calling `initializeConnection` method and do not rethrow that. With having that, there is no information for the user that something went wrong and user is able to click through application.


jrhender (approved):
Makes sense to me, thanks

PR #485, opened 2021-12-01
SWTCH-1257 experimental feature toggle

dawidgil:
Things that need to be done when experimental features are disabled:

- [x] Disable assets page
- [x] Hide off-chain option while enrolling
- [x] Hide Create VC button.
- [x] Hide Assets columns and records connected to assets in enrollment page.

There is also a fix for filtering elements in enrolment view. Previously you could search only by namespace or did. Now those two filters are connected.

### DIFF
--- a/src/app/routes/enrolment/enrolment-list/enrolment-list.component.ts
+++ b/src/app/routes/enrolment/enrolment-list/enrolment-list.component.ts
@@ -1,7 +1,7 @@
 import { Component, Input, OnDestroy, OnInit, ViewChild } from '@angular/core';
 import { ClaimData, NamespaceType, RegistrationTypes } from 'iam-client-lib';
-import { distinctUntilChanged, takeUntil } from 'rxjs/operators';
-import { Subject } from 'rxjs';
+import { take, takeUntil } from 'rxjs/operators';
+import { combineLatest, of, Subject } from 'rxjs';
 import { CancelButton } from '../../../layout/loading/loading.component';
 import { IamService } from '../../../shared/services/iam.service';
 import { LoadingService } from '../../../shared/services/loading.service';
@@ -62,6 +62,10 @@ export class EnrolmentListComponent implements OnInit, OnDestroy {
   ) {
   }
 
+  isAsset(element) {
+    return element?.subject !== element?.claimType && element?.subject !== element?.requester;
+  }
+
   async ngOnInit() {
     // Subscribe to IAM events
     this._iamSubscriptionId = await this.iamService.messagingService.subscribeTo(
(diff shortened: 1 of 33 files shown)


jrhender (approved):
I tested it locally, I think it looks great, thanks @dawidgil !

src/app/routes/enrolment/enrolment-list/enrolment-list.component.ts
```diff
@@ -231,11 +234,12 @@ export class EnrolmentListComponent implements OnInit, OnDestroy {
...
-      .pipe(takeUntil(this._subscription$))
-      .subscribe((reloadList: any) => {
-        if (reloadList) {
-          this.getList(this.dynamicRejected, this.dynamicAccepted);
-        }
+      .pipe(
+        truthy(),
+        takeUntil(this._subscription$)
+      )
+      .subscribe(() => {
```
Malak67, line 241: All subscriptions should be unsubscribed in ngOnDestroy.
As a best practice I would do:

intitialize a variable at the top `subscriptions = new Subscription();`
then wrap each subscription like this `this.subscriptions.add(.....subscribe()..)`
then in ngOnDestroy just do `this.subscriptions.unsubcribe()`

src/app/routes/registration/enrolment-form/enrolment-form.component.spec.ts
```diff
@@ -93,7 +82,8 @@ describe('EnrolmentFormComponent', () => {
       expect(submit.nativeElement.disabled).toBeTruthy();
     });
 
-    it('should have disabled on-chain checkbox when registration types do not contains onChain type', () => {
+    // TODO: probably this test is no longer valid. Need to be checked.
+    xit('should have disabled on-chain checkbox when registration types do not contains onChain type', () => {
```
Malak67, line 86: I would check it and if it not used, just remove it. Otherwise you will forget it.

src/app/routes/enrolment/enrolment-list/enrolment-list.component.ts
Malak67: I just saw you are using takeUntil, that can work as well. But you can take this as an idea.

Malak67 (approved):

src/app/routes/enrolment/enrolment-list/enrolment-list.component.ts
dawidgil: Yeah, both solutions are equivalent. I just prefer and i used to use takeUntil. The only thing that there is needed to have in mind, that takeUntil needs to be last operator otherwise there might be unpredicted behavior.

src/app/routes/registration/enrolment-form/enrolment-form.component.spec.ts
dawidgil: Yeah, you're right! I will remove it.

PR #491, opened 2021-12-03
[SWTCH-1259] On-chain role requester, add an approved role

davycastillo:


jrhender (approved):
Looks good to me, thanks!

PR #492, opened 2021-12-03
feat: add check if role is on chain and is stored in contract.

dawidgil:


src/app/state/pool/pool.effects.ts
```diff
@@ -319,4 +327,12 @@ export class PoolEffects {
         })
       );
   }
+
+  private openNotEnrolledRoleDialog() {
+    this.dialog.open(NotEnroledRoleInfoComponent, {
```
jrhender, line 332: Does the `NotEnroledRoleInfoComponent` have further logic to either redirect you to the role request or to prompt you to add to the `ClaimManager`?

src/app/state/pool/pool.effects.ts
```diff
@@ -264,21 +265,28 @@ export class PoolEffects {
...
-              width: '400px',
-              maxWidth: '100%',
-              disableClose: true,
-            });
-          })
+          filter((roles) => {
+            const length = roles.filter((item) =>
+              item.claimType === REQUIRED_ROLE_FOR_STAKING && item.registrationTypes.includes(RegistrationTypes.OnChain)).length;
+            if (length === 0) {
+              this.openNotEnrolledRoleDialog();
```
jrhender, line 276: Also, a bit pedantic, but I think we've been using `enroled` with one `l` (not `enrolled`) across SB so far
```suggestion
              this.openNotEnroledRoleDialog();
```

src/app/state/pool/pool.effects.ts
dawidgil: This is how it looks like: image
And here is task description: https://energyweb.atlassian.net/browse/DS-100

src/app/state/pool/pool.effects.ts
dawidgil: haha, i will change it. Webstorm all the time is complaining that it should be `enroll` :D

jrhender (approved):
Looks good!

PR #496, opened 2021-12-06
chg(deps): update iam-client-lib to 4.0.0-alpha.1

jrhender:
There's a breaking rename of `chainName` to `chainDisplayName` that isn't yet handled.


dawidgil (approved):
There is no breaking change from switchboard perspective. Iam didn't change any interface, so it is fine to merge.

PR #497, opened 2021-12-06
fix: displaying list of enrolment requests.

dawidgil:


jrhender (approved):
Looks good, I think this is saying that on the main enrolment page, don't show asset enrolment. Is that right? How did the logic end up being wrong?

dawidgil:
Showing assets depends on experimental features. If experimental features are enabled then it should show enrolments to assets. If not, then it should hide it. Actual implementation hides everything that is not asset when experimental feature is enabled -> when switching experimental features it was showing/hiding enrolments to a role not to asset. However before merging i will recheck it.

Malak67 (approved):

PR #501, opened 2021-12-07
chg: remove staking and staking pool.

dawidgil:
Removing entirely staking from switchboard as this is moved to another repo. it was hidden but still it was accessible to visit staking page. After this PR this won't be possible.


jrhender (approved):
I didn't really look to closely at it but it seems like it is all deletions and we'll probably notice if there is a problem.

dawidgil:
Yeah, i just removed all connection to staking and stake ewt pages. The only thing that is also removed is button to launch staking pool (but component for it's creation still exists).

PR #502, opened 2021-12-08
feat(nats): nats env name

Passerino:
NATS_ENVIRONMENT_NAME is required for correct nats events namespace subscribtion

### DIFF
--- a/src/app/shared/services/env/env.service.ts
+++ b/src/app/shared/services/env/env.service.ts
@@ -12,4 +12,5 @@ export class EnvService {
   kmsServerUrl = undefined;
   stakingPoolFactoryAddress = undefined;
   showAzureLoginOption = true;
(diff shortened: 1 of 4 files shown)


src/app/shared/services/env/env.service.ts
```diff
@@ -12,4 +12,5 @@ export class EnvService {
   kmsServerUrl = undefined;
   stakingPoolFactoryAddress = undefined;
   showAzureLoginOption = true;
+  NATS_ENVIRONMENT_NAME = 'ewf-volta';
```
jrhender, line 15: @Passerino is this meant to be used for messaging with `staging` cache-server? If so, can we use `staging` in env variable?
```suggestion
  NATS_ENVIRONMENT_NAME = 'ewf-staging';
```

dawidgil (approved):
Looks good! 👍

PR #503, opened 2021-12-08
chg: use localstorage to handle user data

dawidgil:
https://energyweb.atlassian.net/browse/SWTCH-1275


jrhender (approved):
I still think this feature is a bad idea 😞 . Just private data to be swiped by XSS. But I think the implementation looks fine 😄

dawidgil:
Hopefully we will implement it correctly in near feature. Thanks for checking!

PR #505, opened 2021-12-09
feat: possibility to change claim manager contract address.

dawidgil:


src/app/shared/services/iam.service.ts
```diff
@@ -237,11 +238,15 @@ export class IamService {
...
   }
 
-  private getChainConfig() {
-    const chainConfig: any = {
+  private getChainConfig(): Partial<ChainConfig> {
+    const chainConfig: Partial<ChainConfig> = {
       rpcUrl: this.envService.rpcUrl,
     };
 
+    if (this.envService.claimManagerAddress) {
```
Malak67, line 246: Will this.envService be always available or it makes sens to add this.envService?.claimManagerAddress ?

Malak67 (approved):

src/app/shared/services/iam.service.ts
dawidgil: It will be always available. This is static js file for configuration purposes. The plan is to build once, deploy everywhere and just change properties in env.js file. This file is loaded when application is initializing. So if this file won't be available, then switchboard will not work.

dawidgil: To have better overview: https://www.jvandemo.com/how-to-use-environment-variables-to-configure-your-angular-application-without-a-rebuild/

src/env.js
```diff
@@ -12,6 +12,7 @@
     window.__env.natsServerUrl = 'https://identityevents-dev.energyweb.org/';
     window.__env.kmsServerUrl = undefined;
     window.__env.stakingPoolFactoryAddress = '0xd61Ac65c0fECfb83108f6E374a57FfF9c692901d';
+    window.__env.claimManagerAddress = '';
```
jrhender, line 15: This is the address for `dev`. How does this get configured for `staging`/`volta`
```suggestion
    window.__env.claimManagerAddress = '0xC3dD7ED75779b33F5Cfb709E0aB02b71fbFA3210';
```

dawidgil: It should be done by devops team -> i need to add description how to handle that. Otherwise if we change a contract -> we need to rebuild and redeploy entire switchboard. With above you just change property in env.js file. Read this one as i mentioned above: https://www.jvandemo.com/how-to-use-environment-variables-to-configure-your-angular-application-without-a-rebuild/

jrhender: ClaimManager addresses:
`dev` : `0xC3dD7ED75779b33F5Cfb709E0aB02b71fbFA3210`
`staging`: `0x561785174DF7f564f2591bA52B253c0F663427aB`

jrhender (approved):

PR #513, opened 2021-12-10
Update env for staging

dawidgil:


scripts/stg-env.js
```diff
@@ -6,8 +6,8 @@ fs.readFile(envConfig, 'utf8', function (err, data) {
         return console.log(err);
     }
     const result = data
-        .replace('cacheServerUrl = \'https://identitycache-dev.energyweb.org/v1\'', 'cacheServerUrl = \'https://volta-identitycache.energyweb.org/v1\'')
-        .replace('natsServerUrl = \'https://identityevents-dev.energyweb.org/\'', 'natsServerUrl = \'https://volta-identityevents.energyweb.org/\'')
+        .replace('cacheServerUrl = \'https://identitycache-dev.energyweb.org/v1\'', 'cacheServerUrl = \'https://identitycache-staging.energyweb.org//v1\'')
```
jrhender, line 9: Is that additional slash correct?
```suggestion
        .replace('cacheServerUrl = \'https://identitycache-dev.energyweb.org/v1\'', 'cacheServerUrl = \'https://identitycache-staging.energyweb.org/v1\'')
```

scripts/stg-env.js
```diff
@@ -6,8 +6,8 @@ fs.readFile(envConfig, 'utf8', function (err, data) {
...
+        .replace('natsServerUrl = \'https://identityevents-dev.energyweb.org/\'', 'natsServerUrl = \'wss://identityevents-staging.energyweb.org\'')
```
jrhender, line 10: Are you pretty sure this change to `wss` from `https` is correct?

jrhender (commented):
A couple questions

scripts/stg-env.js
dawidgil: probably not. I will remove it.

scripts/stg-env.js
dawidgil: Nope, i used value the you provided. I can check it.

jrhender: Ah haha, yea, probably. Idk, I think better to use `https` even though it's weird and then change all the config to `wss` later
```suggestion
        .replace('natsServerUrl = \'https://identityevents-dev.energyweb.org/\'', 'natsServerUrl = \'https://identityevents-staging.energyweb.org\'')
```

dawidgil:
No longer relevant. We switched back to angular environments for now.

PR #533, opened 2021-12-28
Swtch 1276 improve validation

dawidgil:


src/app/routes/applications/new-role/new-role.component.ts
```diff
@@ -244,9 +243,9 @@ export class NewRoleComponent implements OnInit, AfterViewInit {
     return this.roleForm.get(control).hasError(errorType);
   }
 
-  alphaNumericOnly(event: any, includeDot?: boolean) {
+  alphaNumericOnly() {
     this.isExistsRoleName = false;
-    return isAlphaNumericOnly(event, includeDot);
+    return true;
```
Malak67, line 248: why do you need return true here? Btw it would be nice to add return type to functos

Malak67 (approved):

src/app/routes/applications/new-role/new-role.component.ts
dawidgil: If it would return false, then it would prevent from setting any value into input. Anyway i removed return value.

Btw. this is a workaround for role creation. This part is refactored [here](https://github.com/energywebfoundation/switchboard-dapp/pull/504), but i wanted to have it working right now as expected.

Malak67: Got it

PR #534, opened 2021-12-30
chg: load list of roles after successful login. Display information i…

dawidgil:
…n tooltip why create vc is disabled.

### DIFF
--- a/src/app/routes/enrolment/enrolment.component.html
+++ b/src/app/routes/enrolment/enrolment.component.html
@@ -4,8 +4,10 @@
       <div class="col-xl-12">
         <div class="d-flex flex-column flex-md-row justify-content-between align-items-center">
           <h4 class="color-white strong my-4">Issue Role</h4>
-          <button *ngIf="isExperimentalEnabled$ | async" mat-raised-button
-                  class="btn btn-primary btn-small mb-3 mb-lg-0" (click)="createVC()">Create VC
+          <button *ngIf="(isExperimental$ | async) && issuesRoles" mat-raised-button
+                  class="btn btn-primary btn-small mb-3 mb-lg-0"
+                  (click)="createVC()">
+            Create VC
           </button>
         </div>
         <div class="card card-default mb-3 pb-0">
--- a/src/app/routes/enrolment/enrolment.component.ts
+++ b/src/app/routes/enrolment/enrolment.component.ts
@@ -9,6 +9,7 @@ import { MatDialog } from '@angular/material/dialog';
 import { NewIssueVcComponent } from '../../modules/issue-vc/new-issue-vc/new-issue-vc.component';
 import { Store } from '@ngrx/store';
 import { SettingsSelectors } from '@state';
+import { IssuanceVcService } from '../../modules/issue-vc/services/issuance-vc.service';
 
 @Component({
   selector: 'app-enrolment',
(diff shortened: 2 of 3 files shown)


src/app/routes/enrolment/enrolment.component.ts
```diff
@@ -36,7 +38,8 @@ export class EnrolmentComponent implements AfterViewInit {
   };
 
   public isMyEnrolmentShown = false;
-  isExperimentalEnabled$ = this.store.select(SettingsSelectors.isExperimentalEnabled);
+  isCreateVCVisible$ = combineLatest([this.store.select(SettingsSelectors.isExperimentalEnabled), this.store.select(RoleSelectors.hasRoles)])
+    .pipe(map(([isExperimental, hasRoles]) => isExperimental && hasRoles));
```
jrhender, line 42: I think this is either wrong or the name can be improved. Does `hasRoles` mean that the DID controls some role definitions? Or does it mean that the DID is an issuer of some roles?

jrhender (changes requested):

jrhender:
Also, no so important, but what does `chg` mean in the commit message? Is this a standard commit prefix?

src/app/routes/enrolment/enrolment.component.ts
dawidgil: hasRole selectors returns true/false depending if the user have some roles inside role management tab inside governance. Probably i should update name for example `issuerHasRoles`

dawidgil:
chg = change. I know i do not always follow conventional commits/ angular commit message format. However most of the time i'm squashing commits into one so those smaller are commits "in progress" and mostly they are for me.

src/app/routes/enrolment/enrolment.component.ts
jrhender: Ok thanks. I think this behaviour is incorrect then. I don't think is relevant whether or not the DID controls roles, the relevant consideration is whether or not the DID can issue any roles.

Regarding the variable name, I think a better name would be `controlsRoles`

dawidgil: Ok, got it. Now i know that i used wrong method. I think i should use `getAllowedRolesByIssuer` from domains service. I will adjust it in a moment. Thanks!

dawidgil:
Now it should be correct. I'm using `getAllowedRolesByIssuer` to get list of possible roles that user can issue.

src/app/routes/enrolment/enrolment.component.html
```diff
@@ -4,8 +4,10 @@
       <div class="col-xl-12">
         <div class="d-flex flex-column flex-md-row justify-content-between align-items-center">
           <h4 class="color-white strong my-4">Issue Role</h4>
-          <button *ngIf="isExperimentalEnabled$ | async" mat-raised-button
-                  class="btn btn-primary btn-small mb-3 mb-lg-0" (click)="createVC()">Create VC
+          <button *ngIf="(isExperimental$ | async) && controlsRoles" mat-raised-button
```
jrhender, line 7: Sorry about this. I think my previous comment was pretty confusing. `controlsRoles` was the right name for the previous behaviour. But I think now that the behaviour has been corrected, `issuesRoles` is a better name.
```suggestion
          <button *ngIf="(isExperimental$ | async) && issuesRoles" mat-raised-button
```

jrhender (approved):
Please do the variable rename but then it can be merged. Thanks @dawidgil !

src/app/routes/enrolment/enrolment.component.html
dawidgil: No problem :) Naming sometimes can be hard or tricky... I adjusted PR, thanks for checking!

PR #535, opened 2022-01-03
Remove not needed modules

dawidgil:


src/app/layout/header/dialog-user/dialog-user.component.spec.ts
```diff
@@ -101,6 +101,7 @@ describe('DialogUserComponent', () => {
     fixture.detectChanges();
 
     birthdate.value = todayDate();
+    console.log(birthdate.value);
```
Malak67, line 104: I think you can remove this log

Malak67 (approved):

src/app/layout/header/dialog-user/dialog-user.component.spec.ts
```diff
@@ -123,5 +124,5 @@ const selectors = (hostDebug: DebugElement) => {
 
 const todayDate = () => {
   const date = new Date(Date.now());
-  return `${date.getMonth()}/${date.getDate()}/${date.getFullYear()}`;
+  return `${date.getMonth() + 1}/${date.getDate()}/${date.getFullYear()}`;
```
jrhender, line 127: Is this a bugfix? Or is this a refactor as it was being done elsewhere?

src/app/layout/header/dialog-user/dialog-user.component.spec.ts
dawidgil: yeah, forgot while debugging test. Thanks.

src/app/layout/header/dialog-user/dialog-user.component.spec.ts
dawidgil: No, this is a fix for bug inside a test. `getMonth()` gets value from 0 to 11, not 1 to 12. It was working until now - because getMonth returned value bigger than 0. So it created a valid date. Now it was returning 0/4/2022 - which is wrong.

PR #536, opened 2022-01-04
bug/ICL-171 onchain claim is not registered from mobile

JGiter:
https://energyweb.atlassian.net/browse/ICL-171

### DIFF
--- a/src/app/shared/services/iam.service.ts
+++ b/src/app/shared/services/iam.service.ts
@@ -87,6 +87,10 @@ export class IamService {
     return this.signerService.providerType;
   }
 
+  get isEthSigner() {
+    return this.signerService.isEthSigner;
+  }
+
   issueClaim(data: { subject: string; claim: any }) {
     return this.wrapWithLoadingService(this.claimsService.issueClaim(data));
   }
--- a/src/app/shared/services/login/login.service.ts
+++ b/src/app/shared/services/login/login.service.ts
@@ -2,7 +2,7 @@ import { Injectable } from '@angular/core';
 import { IamService, PROVIDER_TYPE } from '../iam.service';
 import { LoadingService } from '../loading.service';
 import { ToastrService } from 'ngx-toastr';
-import { AccountInfo, ProviderType, PUBLIC_KEY } from 'iam-client-lib';
+import { AccountInfo, ProviderType, PUBLIC_KEY, IS_ETH_SIGNER } from 'iam-client-lib';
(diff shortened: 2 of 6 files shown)


src/app/shared/services/login/login.service.ts
```diff
@@ -45,11 +45,16 @@ export class LoginService {
   }
 
   isSessionActive() {
-    return (Boolean(localStorage.getItem(PROVIDER_TYPE)) && Boolean(localStorage.getItem(PUBLIC_KEY)));
+    return (
+      Boolean(localStorage.getItem(PROVIDER_TYPE)) 
+      && Boolean(localStorage.getItem(PUBLIC_KEY)) 
+      && Boolean(localStorage.getItem(IS_ETH_SIGNER))
```
dawidgil, line 51: is `IS_ETH_SIGNER` const used somewhere inside iam-client-lib? The same with `PUBLIC_KEY` ? It's more like a question, if we need to handle it inside library as this is for storing session.

JGiter: Yes, iam-lib uses both these. `IS_ETH_SIGNER` is used to correctly sign message and `PUBLIC_KEY` is needed by did-reg, but might be deprecated later

src/app/shared/services/iam.service.ts
```diff
@@ -87,6 +87,10 @@ export class IamService {
     return this.signerService.providerType;
   }
 
+  get isEthSigner() {
+    return this.signerService.isEthSigner;
```
dawidgil, line 91: Another question, can this be null or undefined or always is set to true/false ?

JGiter: It should be defined as long as the session is active

dawidgil (commented):
Approved. However tests are failing, so please fix it before merging.

src/app/shared/services/login/login.service.ts
jrhender: @JGiter I think we can remove `PUBLIC_KEY` as we're no longer requiring that it be added to a DID document, right? We can do this in a separate story though

JGiter: @jrhender  That is right, but first we need to change `EwSigner` to not require public key

jrhender: Ah okay, makes sense, thanks @JGiter

dawidgil (approved):

PR #553, opened 2022-01-17
Task/icl 192 add shared configs

Harasz:


cypress/integration/ui/dashboard.spec.ts
```diff
@@ -1,17 +1,27 @@
...
-    }, {fixture: 'claims/requester-claims-basic.json'}).as('requesterClaim');
-    cy.intercept({method: 'GET', url: '**/DID/*'}, {fixture: 'did/cached-did.json'}).as('cachedDidDoc');
+    cy.intercept(
+      { method: 'GET', url: '**/assets/offered_to/*' },
+      { body: {} }
+    ).as('offeredTo');
+    cy.intercept({ method: 'GET', url: '**/claim/issuer/*' }, { body: {} }).as(
+      'issuerClaim'
+    );
+    cy.intercept(
```
dawidgil, line 11: Another way of aligning the same case. This is randomly aligned?

cypress/plugins/index.ts
```diff
@@ -1,4 +1,5 @@
 // Plugins enable you to tap into, modify, or extend the internal behavior of Cypress
 // For more info, visit https://on.cypress.io/plugins-api
-module.exports = (on, config) => {
+module.exports = () => {
+  // pass
```
dawidgil, line 4: What's the purpose of this comment?

patch.js
```diff
@@ -1,11 +1,16 @@
+// eslint-disable-next-line
```
dawidgil, line 1: why disable next line?

patch.js
```diff
@@ -1,11 +1,16 @@
...
 const fs = require('fs');
-const f = 'node_modules/@angular-devkit/build-angular/src/webpack/configs/browser.js';
+const f =
```
dawidgil, line 3: Why this is aligned this way? it's not very long string...

src/app/app.component.ts
```diff
@@ -11,177 +11,257 @@ import { ThemesService } from './core/themes/themes.service';
...
 export class AppComponent implements OnInit {
-
-  constructor(public settings: SettingsService,
-              private matIconRegistry: MatIconRegistry,
-              private domSanitizer: DomSanitizer,
-              private urlHistoryService: UrlService,
-              private themeService: ThemesService,
-              private store: Store) {
+  constructor(
+    public settings: SettingsService,
```
dawidgil, line 18: I prefer previous alignment. And i would like to stick with old one at least in this repo, because I'm the only person that works on switchboard atm.

src/app/core/menu/menu.service.ts
```diff
@@ -1,24 +1,26 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
dawidgil, line 1: Instead of doing this i think you could change `any` into `unknown`.

src/app/layout/components/user-menu/user-did/user-did.component.html
```diff
@@ -1,5 +1,9 @@
-<div class="d-flex align-items-center justify-content-center did-wrapper mx-2 px-3 mb-2"
-     appCopyToClipboard [copyClipboard]="did" message="User Did">
-  <div class="f-14" data-qa-id="menu-did">{{did | didFormatMinifier}}</div>
+<div
+  class="d-flex align-items-center justify-content-center did-wrapper mx-2 px-3 mb-2"
+  appCopyToClipboard
+  [copyClipboard]="did"
+  message="User Did"
+>
```
dawidgil, line 6: I really do not like this change. Why it can't be like this?
```
<div class="d-flex align-items-center justify-content-center did-wrapper mx-2 px-3 mb-2"
     appCopyToClipboard
     [copyClipboard]="did"
     message="User Did">
```
Switchboard is the only frontend/angular app that seems to need use this, why we do not adjust it? If not in general then only in this app?

src/app/layout/header/dialog-user/dialog-user.component.spec.ts
```diff
@@ -57,13 +56,14 @@ describe('DialogUserComponent', () => {
     const data = {
       name: 'name',
       address: '123',
-      birthdate: date.getTime()
+      birthdate: date.getTime(),
     };
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
dawidgil, line 61: I would prefer to disable eslint in spec files. I do not want to mock everything to fit the types or always put this eslint-disable.

dawidgil (changes requested):
I didn't look on every file, but i do not want changes in html for sure like:
```
<a
  class="nav-link d-lg-block d-xl-block cursor-pointer link-footer"
  routerLink="/governance"
>
```
Some changes are bad for me as i prefer different alignments, but i can live with them for now.
Anyway i do not like the way this change is done. It should be done incrementally not everything at once. If it would be done incrementally then we could adjust everything in time, now it is all or nothing.

dawidgil (approved):

PR #556, opened 2022-01-18
fix(tests): randomly failing unit tests in smart search component

dawidgil:


jrhender (approved):
Looks good. Just curious, what is the `smart-search` component?

dawidgil:
This is this component: 
image
It is responsible for searching for roles. It could be reused across application, but for now it is used only in role creation in 2 places afaik.
Probably i could reuse it on dashboard page for searching for orgs/apps/roles but i didn't checked that yet. 
Hmm, i will create a task for reusing it there.

jrhender:
> This is this component: image It is responsible for searching for roles. It could be reused across application, but for now it is used only in role creation in 2 places afaik. Probably i could reuse it on dashboard page for searching for orgs/apps/roles but i didn't checked that yet. Hmm, i will create a task for reusing it there.

Sounds good, thanks!

PR #559, opened 2022-01-21
fix: sentry integration

dawidgil:
Adjusting code to information from sentry docs: https://docs.sentry.io/platforms/javascript/guides/angular/
Creates a file sentry-config.json where is set sentry DSN and later on it is loaded by application. When SENTRY_DSN value exist in this json config, then sentry is initialized.


hejkerooo (approved):

src/environments/environment.ts
```diff
@@ -24,7 +24,7 @@ export const environment = {
   currencyName: 'Volta Token',
   currencySymbol: 'VT',
   blockExplorerUrl: 'https://volta-explorer.energyweb.org',
-
+  SENTRY_ENVIRONMENT: 'localhost',
```
jrhender, line 27: Is there a `localhost` environment for Sentry?

jrhender (changes requested):
We can just pull the DSN from repo env variables, as the DSN can be public https://docs.sentry.io/product/sentry-basics/dsn-explainer/#dsn-utilization https://forum.sentry.io/t/dsn-private-public/6297

src/environments/environment.ts
dawidgil: Not really. I do not know how this works from Sentry side. Those are environments that are showed to me:
image

Should i remove this part? What about develop?

jrhender: Hmm, I think we should perhaps rename `test` to `develop` on Sentry?

src/app/shared/services/sentry/sentry.service.ts
```diff
@@ -0,0 +1,64 @@
...
+
+@Injectable()
+export class SentryService {
+  constructor(private http: HttpClient, private envService: EnvService) {}
+
+  public load() {
+    return this.http
+      .get<{ dsn: string }>('sentry-config.json', {
+        headers: { 'Cache-Control': 'no-store' },
+      })
```
jrhender, line 17:
```suggestion
  /**
     * Storying DSN in github secrets to make it somewhat more difficult to obtain and so less likely that it will be abused.
     * https://github.com/getsentry/sentry-docs/pull/1723/files#r577851784
     * This is why the values is written to a file using the GHA workflow and is loaded here
     */
  public load() {
    return this.http
      .get<{ dsn: string }>('sentry-config.json', {
        headers: { 'Cache-Control': 'no-store' },
      })
```

jrhender (approved):
Better to have DSN be as private as possible, so I think this is a good approach.
Just made a suggestion from some documentation comments regarding this however.

dawidgil:
Yeah, we can rename it. Anyway if we do not specify environment in filter, then we get all errors anyway.

src/app/shared/services/sentry/sentry.service.ts
dawidgil: Added this comment. Thanks! 👍

Harasz (approved):

PR #564, opened 2022-01-24
refactor(ci): move running unit tests to separate workflow

dawidgil:


dawidgil:
I can change it into `npm ci`. Forgot about it, thanks!
Well, at this moment it is in one workflow -> npm i + unit tests + build + deploy for stedin but executed one by one. I wanted to split it, but your solution looks better. Probably i would do something similar -> instead of creating workflow for unit testing i will create workflow for unit testing, linting and building. What do you think?

Harasz:
Thats good idea, these things can run concurrently so it will improve the time of the CI jobs. But remember about shared `node_modules` for jobs

dawidgil:
Adjusted. Please check this again @Harasz 
Anyway it looks like it's working :) Thanks for this proposition!

Harasz (approved):
Looks very good,
That should look every Build&Test workflow in our repositories @jrhender

PR #574, opened 2022-01-31
SWTCH-1265 create a role with an issuer

dawidgil:


src/app/routes/applications/new-role/components/role-name/role-name.component.ts
```diff
@@ -37,10 +37,9 @@ export class RoleNameComponent {
     return this.form.hasError(errorType);
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   alphaNumericOnly(event: any, includeDot?: boolean): boolean {
     this.existAndNotOwner = false;
-    return isAlphaNumericOnly(event, includeDot);
+    return true;
```
Harasz, line 42: Are you sure? :smile:

src/app/routes/applications/new-role/components/role-name/role-name.component.html
```diff
@@ -19,7 +19,7 @@
             type="text"
             placeholder="Role Name"
             required
-            oninput="this.value = this.value.toLowerCase()"
+            oninput="value = value.toLowerCase().replace(/\s+/g, '')"
```
Harasz, line 22: This case is repeating, maybe there is a way to make it more generic across app?

src/app/routes/applications/new-application/new-application.component.html
```diff
@@ -82,7 +82,10 @@ <h4 mat-dialog-title class="mb-0 d-flex justify-content-between align-items-cent
                   *ngIf="appForm.get('appName').value"
                   class="ml-3 card card-result d-flex flex-row justify-content-start align-items-center shadow-none d-flex pt-3 pb-2 mb-3 mr-3">
                   <span class="pb-2">
-                    {{ appForm?.value?.appName + '.' + ENSPrefixes.Application + '.' + appForm?.value?.orgNamespace }}
+                    {{
```
Harasz, line 85: What about 
```suggestion
                    {{ namespace }}
```
Where `namespace` would be updated on service layer?

src/app/routes/applications/new-role/components/role-name/role-name.component.ts
dawidgil: Yeah, i should more focus on making it better than just fixing it. Sorry, i will rewrite more of it :)

Harasz (approved):

PR #575, opened 2022-01-31
chore: update WC provider patches

Harasz:


Harasz:
We have to wait and test it with https://github.com/energywebfoundation/iam-client-lib/pull/445

dawidgil (approved):

PR #579, opened 2022-02-07
feat: implement did contact exchange with scanner and qr code

dawidgil:


Harasz (approved):

src/app/modules/did-book/components/did-book-list/did-book-list.component.ts
```diff
@@ -34,4 +36,14 @@ export class DidBookListComponent {
...
     this.delete.emit(id);
   }
+
+  qrCodeData(element: DidBookRecord): QrCodeData {
+    return {
+      type: ScanType.User,
+      did: element.did,
+      additionalData: {
+        label: element.label,
+      },
```
jrhender, line 46: I think maybe it would be better to keep it more generic in the sense that we can expect every QR code to have a `type` and some `data`, but not necessarily have a `did`, I think
```suggestion
      type: ScanType.User,
      data: {
        did: element.did,
        label: element.label,
      },
```

jrhender (approved):
A small comment, but it looks good to me!

src/app/modules/did-book/components/did-book-list/did-book-list.component.ts
dawidgil: Yeah, good idea. I will adjust code. Thanks!

PR #594, opened 2022-02-11
fix(deps): breaking changes in icl

dawidgil:


src/app/shared/services/iam.service.ts
```diff
@@ -34,6 +35,7 @@ import { finalize, map } from 'rxjs/operators';
 import { EnvService } from './env/env.service';
 import { ChainConfig } from 'iam-client-lib/dist/src/config/chain.config';
 import { EkcSettingsService } from '../../modules/connect-to-wallet/ekc-settings/services/ekc-settings.service';
+import { IOrganization } from 'iam-client-lib/dist/src/modules/domains/domains.types';
```
Harasz, line 38: @JGiter We should export this type directly from `iam-client-lib`

Harasz (approved):

src/app/shared/services/iam.service.ts
dawidgil: Webstorm is making this import. It looks like just `iam-client-lib` also works, so i will adjust it.

PR #596, opened 2022-02-14
feat: update claim manager address

Harasz:
Update iam-client-lib


src/environments/environment.prod.ts
```diff
@@ -11,7 +11,6 @@ export const environment = {
   natsServerUrl: 'https://identityevents.energyweb.org/',
   kmsServerUrl: 'https://kms.energyweb.org/connect/new',
   ekcUrl: 'https://azure-proxy-server.energyweb.org/api/v1',
-  claimManagerAddress: '0x23b026631A6f265d17CFee8aa6ced1B244f3920C',
```
dawidgil, line 14: Are you sure we should remove this? Is this address used somewhere in ICL?

dawidgil: ok, cool. I think you need to update iam.service to remove setting claimManagerAddress into ICL.

dawidgil (approved):

PR #598, opened 2022-02-15
Update packages

dawidgil:


package.json
```diff
@@ -33,41 +33,41 @@
...
-    "@angular/compiler": "13.1.3",
-    "@angular/core": "13.1.3",
-    "@angular/forms": "13.1.3",
-    "@angular/material": "13.1.3",
-    "@angular/material-moment-adapter": "13.1.3",
-    "@angular/platform-browser": "13.1.3",
-    "@angular/platform-browser-dynamic": "13.1.3",
-    "@angular/router": "13.1.3",
-    "@angular/service-worker": "13.1.3",
+    "@angular/animations": "^13.2.2",
```
Harasz, line 36: Is `^` intentional? We didn't have it previously

Harasz (approved):

package.json
dawidgil: I think both solutions are fine, but i will remove it if you prefer :)

PR #614, opened 2022-03-03
Swtch 1321 localization for customization

dawidgil:


Harasz (approved):

src/locale/messages.json
```diff
@@ -0,0 +1,6 @@
+{
+  "locale": "en-US",
+  "translations": {
+    "2987510875342085677": " Welcome to Energy Web Switchboard "
```
jrhender, line 4: Is this key used somewhere already? How was this key number chosen?
Would it be possible to use a key that a meaning relevant to its usage? Maybe in this case: `welcome-text` for example?

jrhender (commented):
Just a question about the `locale` keys

src/locale/messages.json
dawidgil: This key is generated by Angular and shouldn't be changed manually.
However there is an option to change this key to something more meaningful. Example:

```
<h1 class="font-weight-600 text-center mt-5 mt-md-2 pt-5 pt-md-0 mb-4 pb-2 pb-md-5" 
     i18n='welcome | Some description@@WelcomeID'>
            Welcome to Energy Web Switchboard
</h1>
```
which will produce json like this:
`  "translations": {
    "WelcomeID": " Welcome to Energy Web Switchboard "
  }`
References: 
https://angular.io/guide/i18n-common-prepare#i18n-metadata-for-translation
https://angular.io/guide/i18n-common-translation-files#translate-each-translation-file

dawidgil: @jrhender sure, adjusted.

jrhender (approved):
Looks good, thanks!

PR #628, opened 2022-03-23
feat: possibility to add to DID doc enrolment in enrol page, set onChain and offChain as default

dawidgil:
1. Remove Public/Private checkboxes in enrol page -> by default both are used.
2. Remove `Add to ClaimManager` button
3. Add publish button for approved but not synced roles on enrol page and when creating a VC (near approved precondition).
4. There is a copy paste of code -> `enrolment-list.service` will be removed in future as atm 'm getting Circular Dependency error. I'm working on it on different branch to move to store all notifications -> not using it by service, but dispatch messages with store.


Harasz (approved):

Harasz (changes requested):
Not working

Harasz (approved):

PR #635, opened 2022-04-20
Verifiable presentation

whitneypurdum:
- Adds Verfiable Presentation UI (Dawid Gil)
- Intercepts and decodes _oob params (for review) https://energyweb.atlassian.net/jira/software/projects/SWTCH/boards/54?selectedIssue=SWTCH-1309


src/app/routes/dashboard/dashboard.component.ts
```diff
@@ -62,6 +64,12 @@ export class DashboardComponent implements AfterViewInit {
   }
 
   ngAfterViewInit(): void {
+    this.activeRoute.queryParams.subscribe(async (params: any) => {
```
whitneypurdum, line 67: @dawidgil Is this wrong how I implemented this URL param interception? Felt like if I went right to auth guard and added the /vp route, it always re-routed back to dashboard, and the return URL was not correct. LMK how you would handle this better!

src/app/routes/routes.ts
```diff
@@ -42,12 +42,26 @@ export const routes = [
             (m) => m.SearchResultModule
           ),
       },
+      {
+        path: 'vp',
+        loadChildren: () =>
```
whitneypurdum, line 47: Was originally thinking I should add vp to auth guard children, but maybe not if it is being routed to that after the dashboard component (which is authguarded) already loads?

src/app/routes/dashboard/dashboard.component.ts
```diff
@@ -48,7 +49,8 @@ export class DashboardComponent implements AfterViewInit {
     private activeRoute: ActivatedRoute,
     private loadingService: LoadingService,
     private fb: FormBuilder,
-    private store: Store
+    private store: Store,
+    private urlService: UrlService
```
dawidgil, line 53: urlService is not needed here. It is mostly created for back navigation (button to navigate to the previous page depending on what page it was).

src/app/routes/dashboard/dashboard.component.ts
dawidgil: The problem here is with race condition. You have 2 things that are going on at once:
1. Checking url param and navigating - which is fast.
2. Reinitialization authentication - which is slow because it has to call backend endpoints.
The best would be to not care about authentication here. Just navigate to the `vp` route and let that page handle everything -> like in enrol page. When you are not logged in, then it displays dialog with options to login. If you're logged in then open `vp` page or try to reinitialize auth when state is not stored in app. Example how it looks like: `https://switchboard-dev.energyweb.org/enrol?app=test.apps.dawidgil.iam.ewc&roleName=user`. For now you can just copy paste it, as this is very big component and i didn't refactored it fully yet :/

dawidgil, line 67: When you're using observables you do not need to use async/await. FYI Observables uses Observer pattern. Anyway you can do this:
```
this.activeRoute.queryParams.pipe(
  filter(params: {_oob?: string}) => Boolean(_oob))
  map(params: {_oob?: string}) => params._oob),
  take(1) // just to take it only once and unsubscribe, because we might introduce memory leak.
  // I think it should work, but didn't tested it :) 
).subscribe((oob: string) => this.goToVerifiablePresentation(oob)) 
```

src/app/routes/routes.ts
dawidgil: No, there is no need as i explained above. It's better if this will work like enrol page -> check for component `request-claim`. To be more precisely you might copy paste this method: https://github.com/energywebfoundation/switchboard-dapp/blob/develop/src/app/routes/registration/request-claim/request-claim.component.ts#L155

src/app/routes/verifiable-presentation/verifiable-presentation/verifiable-presentation.component.ts
```diff
@@ -1,16 +1,42 @@
...
-  constructor() { }
-
-  ngOnInit(): void {
-  }
-
+    constructor(
+        private route: ActivatedRoute,
+        private loadingService: LoadingService
+    ) { }
+    isAutolistLoading = {
```
dawidgil, line 20: i think this is not needed

src/app/routes/verifiable-presentation/verifiable-presentation/verifiable-presentation.component.ts
```diff
@@ -1,16 +1,42 @@
...
+        value: false,
+      };
+        //logic to decode the presentation
+        ngOnInit(): void {
+            this._initSearch()
+    
+        }
+ 
+    private _initSearch() {
+        this.route.queryParams.subscribe(async (params: any) => {
```
dawidgil, line 31: async is not needed, because you do not use await inside.

src/app/routes/verifiable-presentation/verifiable-presentation/verifiable-presentation.component.ts
```diff
@@ -1,16 +1,42 @@
...
+            this.loadingService.show()
```
dawidgil, line 32: If there is if statement inside observable (without else) then i prefer just to filter it out. Parsing you can move to `map` operator (first you need to use `pipe()` method of observable). if you are afraid of debugging it, then there is an operator `tap` that do not change object, but you can check what is going on inside or do other stuff like displaying loader :) For hiding i'm using `finalize` operator, which works like `finally` in try-catch block.

src/app/routes/dashboard/dashboard.component.ts
```diff
@@ -151,6 +158,12 @@ export class DashboardComponent implements AfterViewInit {
     this.route.navigate(['assets']);
   }
 
+  goToVerifiablePresentation(oob: string) {
+    this.route.navigate(['vp'], {
```
dawidgil, line 162: When you rebase or merge develop into this branch, then there is `RouterConst` object. It would be good to use it instead of string.

dawidgil (commented):
From code perspective i think everything works fine. Just added some clarifications and hints ;) Do you want from me to rebase this branch? Because it is not up to date with develop

whitneypurdum:
This is for the VP UI (done by Dawid) as well as for this.loadingService.hide() - https://energyweb.atlassian.net/jira/software/projects/SWTCH/boards/54?selectedIssue=SWTCH-1309

dawidgil (approved):
please use node in version 16.10 with npm in version 7 (we use package-lock.json file in version 2). We agreed in IAM stack that we are using npm in version 7. If higher version of node/npm support package-lock in version 2 then it's ok. However i do not know how webpack with angular cli will react.

Before merging you need to rebase/merge latest develop branch as there are some conflicts.

Prettier and unit tests probably won't let you merge this, so you will need to remove all empty methods and probably not used code.
With tests just change `describe` -> `xdescribe` those that are failing. It will disable them. I can later add unit tests :)

whitneypurdum:
@dawidgil is there any way to auto apply lint fixes? I'm getting warnings for a lot of files, but not sure what the linting issue is. Also getting errors for the presentation components that the constructors and ngOnInits are empty - how do you handle those errors? I think removing the lifecycle methods even if they aren't empty is not an option. 
```
  constructor() { }

  ngOnInit(): void {
  }
```

PR #644, opened 2022-05-06
fix(sync claim): check for synced on chain to determine registration …

whitneypurdum:
https://energyweb.atlassian.net/jira/software/projects/SWTCH/boards/54/backlog?selectedIssue=SWTCH-1338
- When publishing (syncing claim to DID doc), first check that it is not already registered on chain. 
- This avoids a re-registering on chain if a user has accepted the transaction to sync a claim on chain, rejected the transaction to sync off-chain, then re-attempts the publish process again. 

https://energyweb.atlassian.net/jira/software/projects/SWTCH/boards/54/backlog?selectedIssue=SWTCH-1338

### DIFF
--- a/src/app/routes/enrolment/enrolment-list/enrolment-list.component.ts
+++ b/src/app/routes/enrolment/enrolment-list/enrolment-list.component.ts
@@ -247,17 +247,15 @@ export class EnrolmentListComponent implements OnInit, OnDestroy {
         this.getList(this.dynamicRejected, this.dynamicAccepted);
       });
   }
-  //the method to checke element
   async addToDidDoc(element: EnrolmentClaim) {
-      console.log("in add to did doc method");
     const isRegisteredOnChain = await this.publishRoleService.checkForNotSyncedOnChain(element);
     const {notSyncedOnChain} = isRegisteredOnChain;
-    console.log(notSyncedOnChain, "NOT SYNCED ON CHAIN")
+    // If the element is already synced on chain, only off-chain registration is needed:
     const registrationTypes = notSyncedOnChain ? element.registrationTypes : [RegistrationTypes.OffChain];
     this.publishRoleService
       .addToDidDoc({
         issuedToken: element.issuedToken,
-        registrationTypes: registrationTypes,
+        registrationTypes,
         claimType: element.claimType,
       })
       .pipe(truthy())
--- a/src/app/routes/enrolment/enrolment-list/enrolment-list.component.ts
+++ b/src/app/routes/enrolment/enrolment-list/enrolment-list.component.ts
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { Component, Input, OnDestroy, OnInit, ViewChild } from '@angular/core';
-import { NamespaceType } from 'iam-client-lib';
+import { NamespaceType, RegistrationTypes } from 'iam-client-lib';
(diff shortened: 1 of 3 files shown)


jrhender (commented):
@whitneypurdum Would it be possible to add tests for this case? I'm not sure it makes sense, I think you might need to mock `publishRoleService` and spy on the call to `addToDidDoc`. 
@dawidgil Do you think it makes sense to add tests in this case?

src/app/routes/enrolment/enrolment-list/enrolment-list.component.ts
```diff
@@ -247,12 +247,18 @@ export class EnrolmentListComponent implements OnInit, OnDestroy {
         this.getList(this.dynamicRejected, this.dynamicAccepted);
       });
   }
-
-  addToDidDoc(element: EnrolmentClaim) {
+  async addToDidDoc(element: EnrolmentClaim) {
```
dawidgil, line 250: This logic should be done in `PublishRoleService`, not in enrolment-list component, because we have 3 places from where we can publish role. This solves only one place, but all of them are using publish-role.service.ts file.
All places from where you can publish role:
1. Enrol page
2. Creating a VC
3. My enrolments list,

and all of them are using `PublishRoleService`.

src/app/shared/services/publish-role/publish-role.service.ts
```diff
@@ -115,25 +116,33 @@ export class PublishRoleService {
...
     registrationTypes: RegistrationTypes[];
     claimType: string;
+    claimTypeVersion: string;
   }) {
+    const isRegisteredOnChain = await this.checkForNotSyncedOnChain(element);
+    const { notSyncedOnChain } = isRegisteredOnChain;
+    // If the element is already synced on chain, only off-chain registration is needed:
+    const registrationTypes = notSyncedOnChain
+      ? element.registrationTypes
+      : [RegistrationTypes.OffChain];
```
whitneypurdum, line 130: Ah, thanks @dawidgil that makes sense. I moved this logic to publish-role service here. LMK if you think this makes better sense. If correct, I will update failing unit tests for toastr. Thank you!

src/app/routes/enrolment/enrolment-list/enrolment-list.component.ts
whitneypurdum: Thanks @dawidgil see below where I moved to publish role service.

src/app/shared/services/publish-role/publish-role.service.ts
```diff
@@ -115,25 +116,33 @@ export class PublishRoleService {
...
-        registrationTypes: element.registrationTypes,
+        registrationTypes: registrationTypes,
         claim: {
           token: element.issuedToken,
           claimType: element.claimType,
         },
       })
       .pipe(
         map((retVal) => {
+            console.log(retVal, "THE RET VAL@@@@")
```
dawidgil, line 145: console log :)

src/app/shared/services/publish-role/publish-role.service.ts
dawidgil: Cool. I just checked what's wrong with tests. I had to change async/await code into rxjs like and they are passing. If you want to play with rxjs here, then i can just push unit tests (added one locally that checks registrationTypes). If you want entire solution for this, i can push that as well :)

dawidgil (approved):
Lgtm

PR #645, opened 2022-05-10
SWTCH-1337 notifications not updated

dawidgil:
Now Notifications uses the same list of enrolments as lists.
I've created a generic table which i want to reuse for every displayed table. I want also add there filtering (filter by dropdown or input value), by defining filter function. However i will do it when i get some more time to handle that.


src/app/state/enrolments/owned/owned.selectors.spec.ts
```diff
@@ -0,0 +1,72 @@
+// import * as OwnedSelectors from './owned.selectors';
+//
+// describe('Owned Assets Selectors', () => {
+//   const asset: any = {
```
Harasz, line 4: Why is it commented?

dawidgil: Added few tests to owned enrolments and requested enrolments selectors.

Harasz (approved):

PR #653, opened 2022-06-02
fix(revocation): revoke function to use proper namespace

dawidgil:


Harasz (approved):
`claimId` should also work now

PR #657, opened 2022-06-07
#swtch 1317/vp UI

whitneypurdum:
Adds happy path for:
1.  parsing oob params
2. fetching presentation data from ICL using param URL
3. populating UI table (self sign and credential dropdown)
4. Logic to disable the submit button

https://www.loom.com/share/125d6c8fa4c6415f939f46ac1bb667bc


package.json
```diff
@@ -47,22 +47,27 @@
     "@angular/platform-browser-dynamic": "13.2.2",
     "@angular/router": "13.2.2",
     "@angular/service-worker": "13.2.2",
+    "@ew-did-registry/credentials-interface": "^0.6.3-alpha.596.0",
```
dawidgil, line 50: i think this is not needed. Those libs are added by ICL, i don't think we need to care about them here.

src/app/routes/verifiable-presentation/received-presentations/received-presentations.component.ts
```diff
@@ -1,15 +1,25 @@
...
   templateUrl: './received-presentations.component.html',
   styleUrls: ['./received-presentations.component.scss'],
 })
 export class ReceivedPresentationsComponent {
-  displayedColumns: string[] = ['namespace', 'reason'];
-  dataSource = new MatTableDataSource([
-    { namespace: 'installer', reason: 'asdf dsa' },
-    { namespace: 'installer2', reason: 'Some other reason' },
-  ]);
+  @Input() tableData;
```
dawidgil, line 9: It's better to have MatTableData object in the same component that displays it. You can pass here regular array and change it to the MatTableData type. There are 2 options:
1. In ngOnInit assign it. -> ngOnInit() { if(tableData) { ... }
2. create a setter -> @Input set tableData(value: SomeInterface) { ... }
This way it will be easier to test it, you do not care about creating MatTableData object in tests, you're just passing array.

src/app/routes/verifiable-presentation/received-presentations/received-presentations.component.ts
```diff
@@ -1,15 +1,25 @@
...
 })
 export class ReceivedPresentationsComponent {
-  displayedColumns: string[] = ['namespace', 'reason'];
-  dataSource = new MatTableDataSource([
-    { namespace: 'installer', reason: 'asdf dsa' },
-    { namespace: 'installer2', reason: 'Some other reason' },
-  ]);
+  @Input() tableData; 
+  @Input() submitDisabled: boolean;
+  @Input() credentialSelected
```
dawidgil, line 11: If possible please add interfaces.

src/app/routes/verifiable-presentation/verifiable-presentation/verifiable-presentation.component.ts
```diff
@@ -26,34 +35,35 @@ export class VerifiablePresentationComponent implements OnInit {
...
-                } finally {
-                    this.loadingService.hide();
-                }
-                 */
+         TO DO:
+         - Error handling - no url parsed
+         - UI Handling - no results found from initiate exchange
+        */
+        console.log(url);
+        const result = await this.iamService.verifiableCredentialsService.initiateExchange({ type: 'https://energyweb.org/out-of-band-invitation/vc-api-exchange', url });
```
dawidgil, line 60: move getting and formatting all this data to separate service. It will be easier to test because you do not need to mock queryParams. This method can return object like: 
`{ challenge: ..., interact: ..., presDef: ..., data: ...}`.

dawidgil (approved):

package.json
whitneypurdum: I think we need to import these to import the types/interfaces

src/app/routes/verifiable-presentation/verifiable-presentation/verifiable-presentation.component.ts
whitneypurdum: Oh I see you mean like asset.service.ts as an example. got it!

src/app/routes/verifiable-presentation/received-presentations/received-presentations.component.ts
whitneypurdum: Thanks for this!

src/app/routes/verifiable-presentation/received-presentations/received-presentations.component.ts
```diff
@@ -1,15 +1,26 @@
...
+  @Input() set tableData(data: ICredentialTableData[]) {
+    this.dataSource = new MatTableDataSource<ICredentialTableData>(data);
+  }
+  @Input() submitDisabled: boolean;
+  @Input() requiredCredentials: { [key: string]: IVerifiableCredential };
+  displayedColumns: string[] = ['descriptor', 'verification'];
+  dataSource = new MatTableDataSource<unknown>([]);
+
+  handleCredentialUpdate(data: MatSelect) {
+    console.log(data, 'what is this?');
```
dawidgil, line 23: I think this console.log might be removed.

src/app/routes/verifiable-presentation/services/presentation.service.ts
```diff
@@ -0,0 +1,80 @@
...
+  fetchPresentation(presentation: PresentationRequest) {
+    return this.iamService
+      .wrapWithLoadingService(
+        this.iamService.verifiableCredentialsService.initiateExchange({
+          type: presentation.type,
+          url: presentation.url,
+        })
+      )
+      .pipe(
+        map((result) => result),
```
dawidgil, line 23: This map is not needed, because you do not change the result.

dawidgil (approved):
In general looks good. Just some console logs that would be good to remove in the end.

PR #661, opened 2022-06-15
feat: add selected dropdown element when clicked

whitneypurdum:
- For New Role --> Set Restrictions --> Role Precondition - when an item is selected from the dropdown, automatically add it, rather than relying on the "add" button

### DIFF
--- a/src/app/shared/components/smart-search/smart-search.component.ts
+++ b/src/app/shared/components/smart-search/smart-search.component.ts
@@ -62,8 +62,10 @@ export class SmartSearchComponent implements AfterViewInit {
 
   autocompleteSelectionHandler(event: MatAutocompleteSelectedEvent) {
     this.selectionHandler(event.option.value);
+    if (this.isAdding) {
+      this.addSelection();
+    }
   }
-
   selectionHandler(namespace?: string) {
     this.selected.emit({
       namespace,
@@ -83,12 +85,12 @@ export class SmartSearchComponent implements AfterViewInit {
     return this.searchText?.value?.trim()?.length > 2;
   }
 
-  addRole(): void {
+  addSelection(): void {
     if (this.searchText.invalid) {
       return;
     }
     this.add.emit({
-      role: this.searchText.value,
+      value: this.searchText.value,
       searchType: this.searchType,
     });
     this.clear();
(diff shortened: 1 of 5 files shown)


whitneypurdum:
@dawidgil if this implementation is okay, I will update unit tests. I added the condition because this component is shared across the app, so didn't want to force the behavior adoption of adding when dropdown selected.

src/app/shared/components/smart-search/smart-search.component.ts
```diff
@@ -62,8 +63,14 @@ export class SmartSearchComponent implements AfterViewInit {
 
   autocompleteSelectionHandler(event: MatAutocompleteSelectedEvent) {
     this.selectionHandler(event.option.value);
+    if (this.addOnSelect) {
+      this.add.emit({
```
dawidgil, line 67: Here you can reuse `addRole` method. It's the same code, of course inside if statement.

src/app/shared/components/smart-search/smart-search.component.ts
```diff
@@ -26,6 +26,7 @@ import { IOrganization } from 'iam-client-lib/dist/src/modules/domains/domains.t
 export class SmartSearchComponent implements AfterViewInit {
   @Input() searchText: FormControl = new FormControl('');
   @Input() placeholderSearch: string;
+  @Input() addOnSelect: boolean;
```
dawidgil, line 29: I suppose you do not need to add this property. You can use `isAdding` getter. It's checking already if it is for adding role or not. However i see that those names do not describes what they are doing. What do you think about renaming them to be more descriptive?
I see two options:
1. Rename `add -> addRole`
2. change `add` interface, by changing `role` into `value`. Method `addRole -> addSelection` This way we can `add` organization/application by using smart search. 
What do you think?
btw. what do you think about removing `add` button from html?

whitneypurdum: @dawidgil it looks like above it says to change add to addRole, then change addRole to addSelection. Can you clarify? I'm up for removing the add button if everyone is okay with it!

whitneypurdum: @dawidgil Oh NM, I totally see what you mean!

dawidgil (approved):
Looks good, thanks!

PR #669, opened 2022-06-29
fix(role-creation): incorrectly displaying list of issuers and revokers

dawidgil:


jrhender (approved):
Looks good to me, thanks!

PR #681, opened 2022-07-14
Expiration date

dawidgil:


src/app/routes/enrolment/view-requests/components/expiration-date/expiration-date.component.spec.ts
```diff
@@ -98,19 +102,23 @@ describe('ExpirationDateComponent', () => {
...
-      return now.getUTCSeconds() + now.getUTCMinutes() * 60 + (now.getUTCHours()) * 60 * 60 - now.getTimezoneOffset() * 60;
-    }
+      return (
+        now.getUTCSeconds() +
+        now.getUTCMinutes() * 60 +
+        now.getUTCHours() * 60 * 60 -
+        now.getTimezoneOffset() * 60
+      );
+    };
     console.log(Date.now() / 1000);
```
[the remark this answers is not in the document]
dawidgil: No, good catch! Thanks. It was only for testing purposes. I didn't know why my test was failing, so i started writing logs in a test to catch what is wrong.

nadia-adila (approved):

src/app/state/governance/role/services/role.service.ts
```diff
@@ -14,4 +14,12 @@ export class RoleService {
...
     );
   }
+
+  getDefinition(namespace: string) {
+    return this.iamService.getRolesDefinition(namespace);
+  }
+
+  getDefinitions(namespaces: string[]) {
+    return this.iamService.cacheClient.getRolesDefinition(namespaces);
+  }
```
jrhender, line 24: I think this is a bit confusing to someone reading the code. It's a bit unclear why one is using method in `iamService` and one is going to the `cacheClient` directly. To me, it might be clearer to change `IamService.getRolesDefinition` to accept and return an array and then do:

```suggestion
  getDefinition(namespace: string) {
    return this.iamService.getRolesDefinition([namespace])[0];
  }

  getDefinitions(namespaces: string[]) {
    return this.iamService.getRolesDefinition(namespaces);
  }
```

src/app/routes/enrolment/view-requests/components/expiration-date/expiration-date.component.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+    jasmine.clock().uninstall();
+  });
+
+  it('should create', () => {
+    fixture.detectChanges();
+    expect(component).toBeTruthy();
+  });
+
+  // TODO: passing locally, failing on GHA. Check it.
+  xit('should set expiration date and reset it', () => {
```
jrhender, line 59: Is this something you want to do in this PR or should we make a separate task to investigate and fix?

jrhender (approved):
Left a couple comments but it looks good and I think we could merge as is.

src/app/state/governance/role/services/role.service.ts
dawidgil: I will remove method from iamService and i will just use directly from cacheClient.
Anyway getRolesDefinition returns an object, where a key is a role namespace.

src/app/routes/enrolment/view-requests/components/expiration-date/expiration-date.component.spec.ts
jrhender: Ok, sounds good. Could you add a new task to SB backlog so that we don't forget please?

PR #683, opened 2022-07-15
fix: checking if off-chain credential can be revoked

dawidgil:


jrhender (approved):

PR #688, opened 2022-07-19
feat(role-creation): use did book list when setting issuers/revokers

dawidgil:


src/app/modules/did-book/components/select-did/select-did.component.ts
```diff
@@ -1,54 +1,104 @@
...
-  didBook$;
-  newOwnerDID = new FormControl('', [
-    Validators.required,
-    HexValidators.isDidValid(),
-  ]);
-  isNotKnownDid: boolean;
+  didBook$: Observable<DidBookRecord[]>;
+  newDID = new FormControl('', [HexValidators.isDidValid()]);
+  isNotKnownDid: boolean = false;
+  private _didToRemove: string[] = [];
```
jrhender, line 36: Sometimes I've seen that private class fields should come before public ones when being declared. But I think this is a hold over from my C# days and not applicable to Angular 😄 https://stackoverflow.com/a/310967

src/app/modules/did-book/components/select-did/select-did.component.ts
```diff
@@ -1,54 +1,104 @@
...
+    }
+    this.newDID.setValidators([
+      Validators.required,
+      ...this.defaultValidators(),
+    ]);
+  }
+
+  approveHandler(): void {
+    this.dialog.open(AddSingleRecordComponent, {
+      width: '550px',
```
jrhender, line 62: Is this kind of width param hardcoded elsewhere in the app? I would think that this might make the app less usable on mobile 🤔 ?

src/app/routes/applications/new-role/components/role-type-did/role-type-did.component.spec.ts
```diff
@@ -44,65 +41,28 @@ describe('RoleTypeDidComponent', () => {
...
-
-    const { didInput } = selectors(hostDebug);
-    didInput.value = 'did:ethr:0x';
-    dispatchInputEvent(didInput);
-    fixture.detectChanges();
-
-    const { matError } = selectors(hostDebug);
-    expect(matError.textContent).toContain('DID format is invalid');
-  });
-
```
jrhender, line 74: Were these test cases not longer relevant? They seem like decent test cases to me still

jrhender (commented):
It's looking good, just some questions

src/app/modules/did-book/components/select-did/select-did.component.ts
dawidgil: I do not know if are there any good practices for JS/TS/angular about private/public fields. most of the times I'm writing first public second private. Anyway in the end of the day it doesn't matter in JS world, because all properties are public.

src/app/modules/did-book/components/select-did/select-did.component.ts
dawidgil: No, it looks fine on small viewports. I've checked it. When viewport is smaller than 550px dialog is adjusting. However when viewport is bigger, then it is not wider than 550px.

src/app/routes/applications/new-role/components/role-type-did/role-type-did.component.spec.ts
dawidgil: Well, the logic has been moved to `select-did` component. There we have input with autocomplete that contains DIDs from the DID book. There is also handling validations etc. However i can move test `should display error message when element is not a did` there. The test `should display error message about already existing element` is already moved.

jrhender (approved):

PR #690, opened 2022-07-26
bug(expiration): Date fixes and updates to viewing/editing claim expiration in issuance

whitneypurdum:
This PR:
1. This fixes the error that the 'expiration date is in the past' upon issuance. 
2. Updates verbiage for expiration date on claim issuance (see screenshots, loom videos) - handles claims with default expiration and claims with no default expiration
3. Adds "Restore Default Expiration" button that allows a user to revert to the original default expiration date if they have chosen another date with the date picker. This button is shown ONLY IF the chosen expiration timestamp DOES NOT EQUAL the default validity timestamp (i.e. you had a default validity period and you selected a new date)
4. Logged the dates coming back from ICL once claims are issued, no errors seen
Please see bug for further questions and discussion about expiration flow. 

Scenarios tested (let me know if I'm missing anything):
**Scenario 1:**
1. Create a role with a default validity period
2. Request to enrol
3. On issuance - issue role with no changes to expiration
4. Sync to chain

**Scenario 2:**
1. Create a role with a default validity period
2. Request to enrol
3. On issuance - update claim with new expiration date
4. Sync to chain

**Scenario 3:**
1. Create a role with a default validity period
2. Request to enrol
3. On issuance - update claim with new expiration date
4. **Restore** to original validity period
5. Sync to chain

**Scenario 4:**
1. Create a role with a default validity period
2. Request to enrol
3. On issuance - **remove validity period**
5. Sync to chain
*There is a[ bug on ICL](https://energyweb.atlassian.net/jira/software/projects/ICL/boards/85?selectedIssue=ICL-338) for this scenario that does not allow a user to issue a claim of no exp date if there is a default date associated with the role

**Scenario 5:**
1. Create a role with a default validity period
2. Request to enrol
3. On issuance - **remove validity period**
4. **Restore** to original validity period
5. Sync to chain

**Scenario 6:**
1. Create a role with **NO** default validity period
2. Request to enrol
3. On issuance **create expiration date**
5. Sync to chain

**NO DEFAULT VALIDITY PERIOD ON ROLE:**
https://www.loom.com/share/ef0f7a8027f8481c9d6e0278d3b74e43
[
Screen Shot 2022-07-27 at 7 15 22 PM
](url)

Screen Shot 2022-07-27 at 7 15 11 PM

**ROLE WITH DEFAULT VALIDITY PERIOD:**
https://www.loom.com/share/277827f658b542168b17d130b80db9bc

Screen Shot 2022-07-27 at 7 17 06 PM

Screen Shot 2022-07-27 at 7 16 56 PM

Screen Shot 2022-07-27 at 7 16 44 PM


src/app/routes/enrolment/view-requests/issuer-requests/issuer-requests.component.ts
```diff
@@ -84,7 +84,7 @@ export class IssuerRequestsComponent
   }
 
   updateExpirationDate(expirationTime: number): void {
-    this.expirationTime = expirationTime;
+    this.expirationTime = Date.now() + expirationTime * 1000;
```
jrhender, line 87: I would change this param to `validityPeriod` as I think it is more accurate. 

I think this overall situation is a bit confusing though as the `expirationTime`/`validityPeriod` param is in milliseconds but `this.expirationTime` is in seconds? Maybe it's better to do the `* 1000` before this function so that everything is in seconds for this function?

```suggestion
    this.expirationTime = Date.now() + validityPeriod * 1000;
```

src/app/routes/enrolment/view-requests/components/expiration-info/expiration-info.component.html
```diff
@@ -1,11 +1,12 @@
...
-    Default expiration date of this role is <b>{{ defaultValidityPeriod | timeShift | defaultDate }} UTC</b>.
+    The default expiration date for this role is <b>{{ defaultValidityPeriod | timeShift | defaultDate }} UTC</b>.
   </ng-container>
   <ng-container *ngIf="!defaultValidityPeriod"> Default expiration date is not set. </ng-container>
-  This role will be issued with
+
   <ng-container *ngIf="expirationTime"
-    ><b>{{ expirationTime | timeShift | defaultDate }} UTC</b></ng-container
+    >This role will now be issued with an expiration date of
+    <b>{{ expirationTime | timeShift | defaultDate }} UTC</b></ng-container
```
jrhender, line 9: Does this line make sense to you @dawidgil @whitneypurdum ? Is this the use of the Bitwise OR https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_OR ?

src/app/routes/enrolment/view-requests/components/expiration-info/expiration-info.component.html
```diff
@@ -1,11 +1,12 @@
...
   >
-  <ng-container *ngIf="!expirationTime"><b>no</b></ng-container> expiration date
+  <!-- <ng-container *ngIf="!expirationTime"><b>no</b></ng-container> expiration date -->
```
jrhender, line 11: Why comment this out?

jrhender (commented):
I think it might be a good fix but a few comments/questions

src/app/shared/pipes/time-duration/timestamp.ts
```diff
@@ -25,6 +25,9 @@ export class Timestamp {
   }
 
   parseToSeconds(value: ParseTimestampResult): number {
```
whitneypurdum, line 27: @dawidgil this was sending 0 to ICL, so ICL was storing 0 as default validity period and then SB was then using it as a timestamp. If everything is 0 here, it sets the timestamp to null.

src/app/routes/enrolment/utils/format-validity-period-string.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (!num) {
+    return null;
+  }
+  const seconds = Math.floor(num % 60);
+  const minutes = Math.floor((num % 3600) / 60);
+  const hours = Math.floor((num % (3600 * 24)) / 3600);
+  const days = Math.floor(num / (3600 * 24));
+
+  const unitTimesFormatted = {
+    0: {
```
whitneypurdum, line 11: This can be better

whitneypurdum:
Is it strange that we don't have 'months'? I would think months in this scenario might be as or  more helpful than seconds or minutes? 
Screen Shot 2022-07-27 at 8 16 28 PM

src/app/routes/enrolment/view-requests/components/expiration-info/expiration-info.component.html
whitneypurdum: Yeah, I wasn't sure about it either but it is something with maybe Angular syntax? It converts timestamps to formatted date strings. I'm not sure where it comes from, it took me a moment to realize that it was transforming the timestamps.

src/app/routes/enrolment/view-requests/issuer-requests/issuer-requests.component.ts
whitneypurdum: I updated the param name here, but @dawidgil  I'll let you weigh in on where we should do the multiplication

src/app/routes/enrolment/utils/format-validity-period-string.ts
```diff
@@ -0,0 +1,60 @@
...
+    DAYS = 'days',
+    HOUR = 'hour',
+    HOURS = 'hours',
+    MINUTE = 'minute',
+    MINUTES = 'minutes',
+    SECOND = 'second',
+    SECONDS = 'seconds',
+  }
+
+  const years = {
```
whitneypurdum, line 19: I'm wondering if a library like momentJS already does calculations like this...

src/app/shared/pipes/time-duration/timestamp.ts
dawidgil: Yeah, i didn't think about this case. Thanks!

src/app/routes/enrolment/utils/format-validity-period-string.ts
dawidgil: I think momentJS is not. And i would not use it any more, because it is deprecated and it would be better to remove it and use something new. Anyway there is a pipe `timeDuration` that is handling this -> it's already tested, you can check if it fits your needs.

src/app/routes/enrolment/view-requests/issuer-requests/issuer-requests.component.ts
dawidgil: I think the multiplication should be done in `app-expiration-date` component. At least you can adjust calculation there :)

src/app/routes/enrolment/view-requests/components/expiration-date/expiration-date.component.html
```diff
@@ -1,26 +1,40 @@
 <mat-label class="ml-3 pl-1">Expiration Date</mat-label>
 <app-expiration-info
   [defaultValidityPeriod]="defaultValidityPeriod"
-  [expirationTime]="expirationTimeShift"></app-expiration-info>
+  [expirationTime]="expirationTimeShift"
```
jrhender, line 4: Maybe this a side note to this PR, but does the name `expirationTimeShift` make sense to you here? Seems like a "time shift" is more of a time period than a specific time, so it's like there is a type mismatched here.

dawidgil:
> Is it strange that we don't have 'months'? I would think months in this scenario might be as or more helpful than seconds or minutes? Screen Shot 2022-07-27 at 8 16 28 PM

@whitneypurdum It might be confusing, but how many days have month? It can be 28, 29, 30 and 31. It was hard to pick something. It was just easier and imo more user friendly to ignore months and allow to set days from 0 to 365.
However if you have an idea how to handle this then I'm open for it :)

src/app/routes/enrolment/view-requests/components/expiration-info/expiration-info.component.html
dawidgil: Those are so called pipes: https://angular.io/guide/pipes. They are transforming data in html. It's easier to reuse and test them instead of writing the same code in component class.

jrhender: @dawidgil Thanks for the link!

src/app/routes/enrolment/view-requests/components/expiration-date/expiration-date.component.html
whitneypurdum: @dawidgil

src/app/routes/enrolment/utils/format-validity-period-string.ts
whitneypurdum: I think that is taking the individual values rather than parsing from a timestamp? should I move this logic to that class?

dawidgil: It's parsing from seconds to readable years/days/hours/minutes/seconds text (check tests). No need to move this logic, because that pipe is already doing the same thing. Look for example for role creation where you're setting defaultValidityPeriod.

src/app/routes/enrolment/view-requests/components/expiration-date/expiration-date.component.html
dawidgil: The problem with previous implementation was that I had wrong assumption -> I though expirationTimestamp is described in seconds like defaultValidityPeriod. But expirationTimestamp is a date in miliseconds. So renaming it makes sense.

src/app/routes/enrolment/view-requests/issuer-requests/issuer-requests.component.ts
whitneypurdum: I don't think expirationTime is being passed into that component to do the calculation there? From what I see it stays in issuer-request?

src/app/routes/enrolment/view-requests/components/expiration-date/expiration-date.component.html
jrhender: So sounds like we should rename `expirationTimeShift` to `expirationTime`? Am I understanding this right?

jrhender:
@whitneypurdum could you take a the eslint&prettier job when you have a moment? https://github.com/energywebfoundation/switchboard-dapp/runs/7561703485?check_suite_focus=true

jrhender (approved):
I think this is a good improvement as is, so we should get it in, IMO

jrhender:
Though I think it was good to get this, I was testing a bit and I think there is still a mismatch somewhere 

You can see in the screenshot that I set a default validity period of 2 hours for the role.

I opened the issued role dialog at 20:37 CEST , so the default expiry time should be 22:37 CEST but the dialog says that the expiry will be 22:37 **UTC** (which is 2 hours behind CEST). So maybe there are some timezone issues?

image

whitneypurdum:
I’ll look into this. Maybe time stamps aren’t set in uTC but the html in Angular date pipe is in UTC

PR #691, opened 2022-07-27
fix: displaying preview from owned enrolments

dawidgil:


jrhender (approved):
Looks good, thanks @dawidgil !

dawidgil:
I've added some tests for checking if correct component is used for displaying dialog.

PR #693, opened 2022-07-28
Display raw did document

dawidgil:


src/app/layout/header/header.component.html
```diff
@@ -107,6 +107,18 @@ <h6 class="mb-0">Enrolments</h6>
...
     </button>
+    <ng-container *ngIf="isExperimentalEnabled$ | async">
+      <button
+        class="color-link px-4 pb-2 border-bottom-1 position-relative"
+        mat-menu-item
+        appShowRawData
+        [data]="didDocument$ | async"
+        header="Raw DID Document"
+        data-qa-id="menu-raw-did-document">
+        Raw DID Document
```
jrhender, line 118:
```suggestion
        header="DID Document"
        data-qa-id="menu-did-document">
        DID Document
```

jrhender (approved):
Looks good! In my opinion, we can just saw "DID Document", instead of "Raw DID Document", as I'm not sure what a "non-raw" DID Document would be

src/app/layout/header/header.component.html
dawidgil: @jrhender ok, thanks. I updated PR :)

PR #695, opened 2022-07-29
fix(exp): store defaultValidityPeriod as milliseconds

whitneypurdum:
https://energyweb.atlassian.net/browse/SWTCH-1380

### DIFF
--- a/src/app/routes/enrolment/utils/format-validity-period-string.ts
+++ b/src/app/routes/enrolment/utils/format-validity-period-string.ts
@@ -1,5 +1,0 @@
-import { TimeDurationPipe } from 'src/app/shared/pipes/time-duration/time-duration.pipe';
-export const formatValidityPeriod = (num: number): string => {
-  const timePipe = new TimeDurationPipe();
-  return num ? timePipe.transform(num / 1000) : null;
-};
--- a/src/app/routes/enrolment/utils/format-validity-period-string.ts
+++ b/src/app/routes/enrolment/utils/format-validity-period-string.ts
@@ -1,5 +1,5 @@
 import { TimeDurationPipe } from 'src/app/shared/pipes/time-duration/time-duration.pipe';
 export const formatValidityPeriod = (num: number): string => {
   const timePipe = new TimeDurationPipe();
-  return num ? timePipe.transform(num) : null;
+  return num ? timePipe.transform(num / 1000) : null;
 };
--- a/src/app/routes/enrolment/view-requests/components/expiration-date/expiration-date.component.spec.ts
+++ b/src/app/routes/enrolment/view-requests/components/expiration-date/expiration-date.component.spec.ts
@@ -99,7 +99,7 @@ describe('ExpirationDateComponent', () => {
 
     const { expirationDateInput } = getSelectors(hostDebug);
 
-    expect(new Date(expirationDateInput.value)).toEqual(new Date(2000, 0, 2));
+    expect(new Date(expirationDateInput.value)).toEqual(new Date(2000, 0, 1));
   });
 });
 const getSelectors = (hostDebug) => ({
--- a/src/app/routes/enrolment/view-requests/components/expiration-date/expiration-date.component.spec.ts
+++ b/src/app/routes/enrolment/view-requests/components/expiration-date/expiration-date.component.spec.ts
@@ -94,12 +94,12 @@ describe('ExpirationDateComponent', () => {
   });
 
   it('should set default time in input from current time', () => {
-    component.defaultValidityPeriod = 60 * 60 * 24;
+    component.defaultValidityPeriod = 60 * 60 * 24 * 1000;
     fixture.detectChanges();
 
     const { expirationDateInput } = getSelectors(hostDebug);
 
-    expect(new Date(expirationDateInput.value)).toEqual(new Date(2000, 0, 1));
+    expect(new Date(expirationDateInput.value)).toEqual(new Date(2000, 0, 2));
   });
 });
 const getSelectors = (hostDebug) => ({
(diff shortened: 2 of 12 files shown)


src/app/routes/enrolment/utils/format-validity-period-string.ts
```diff
@@ -1,5 +1,5 @@
 import { TimeDurationPipe } from 'src/app/shared/pipes/time-duration/time-duration.pipe';
 export const formatValidityPeriod = (num: number): string => {
```
dawidgil, line 2: why you're transforming it here instead using pipe in html?

src/app/shared/pipes/time-duration/timestamp.ts
```diff
@@ -37,6 +37,19 @@ export class Timestamp {
     );
   }
 
+  parseToMilliseconds(value: ParseTimestampResult): number {
```
dawidgil, line 40:
```suggestion
    parseToMilliseconds(value: ParseTimestampResult): number {
    const seconds = this.parseToSeconds(value);
    return seconds ? seconds * 1000 : null;
  }
```

Would be good to add some tests as well.

src/app/routes/enrolment/view-requests/components/expiration-info/expiration-info.component.html
```diff
@@ -23,7 +23,7 @@
   <ng-container *ngIf="(!removeExpDate && defaultValidityPeriod) || expirationTime">
     <br />
     This role will now be issued with an expiration date of
-    <b>{{ expirationTime || defaultValidityPeriod | timeShift | defaultDate }} UTC</b>
+    <b>{{ expirationTime / 1000 || defaultValidityPeriod / 1000 | timeShift | defaultDate }} UTC</b>
```
dawidgil, line 26: What do you think about changing timeShift pipe to accept miliseconds instead of seconds? Or we can add optional property which will inform how it should be treated: as seconds or miliseconds. Then this line would like like this:
<b>{{ expirationTime || defaultValidityPeriod | timeShift: 'miliseconds' | defaultDate }} UTC</b>

whitneypurdum:
@dawidgil Any idea why this unit test is failing? I think there is still an issue with expiration code but tbh I can't spot it. Could be in timeshitt calcuation or calchours etc. but I'm pretty stumped about it.

src/app/routes/enrolment/utils/format-validity-period-string.ts
whitneypurdum: How do I just call the transform method in the pipe in the html. I don't want a date, I want the hours, seconds, etc.

dawidgil:
> @dawidgil Any idea why this unit test is failing? I think there is still an issue with expiration code but tbh I can't spot it. Could be in timeshitt calcuation or calchours etc. but I'm pretty stumped about it.

@whitneypurdum This test is failing because you're no longer multiply defaultValidityPeriod by 1000. Earlier it was in seconds, now you're passing miliseconds. Test still assumes that it is seconds.

src/app/routes/enrolment/utils/format-validity-period-string.ts
dawidgil: pipe have a `transform` method which changes variable/property/object into something more readable. It is very easy to test and reuse. Here is link: https://angular.io/guide/pipes
To use this pipe in html all you need is:
`{{ defaultValidityPeriod | timeDuration }}`.

whitneypurdum: Ah sorry, I was looking at time shift pipe! Thanks.

src/app/routes/enrolment/view-requests/components/expiration-date/expiration-date.component.spec.ts
```diff
@@ -99,7 +99,7 @@ describe('ExpirationDateComponent', () => {
 
     const { expirationDateInput } = getSelectors(hostDebug);
 
-    expect(new Date(expirationDateInput.value)).toEqual(new Date(2000, 0, 2));
+    expect(new Date(expirationDateInput.value)).toEqual(new Date(2000, 0, 1));
```
dawidgil, line 102: Is this test checking right now something? I'm not sure, because default date is 1st January 2000, and after setting defaultValidityPeriod it is still 1st January 2000, but with different hour/minute/seconds. To adjust test change defaultValidityPeriod from 1 day in seconds into 1 day in miliseconds.
```
    component.defaultValidityPeriod = 60 * 60 * 24 * 1000;
```
This way we will know that setting defaultValidityPeriod is changing input.

src/app/routes/enrolment/view-requests/components/expiration-info/expiration-info.component.html
dawidgil: You can leave it as it is right now. Anyway if we go with miliseconds everywhere, we should late adjust it to handle miliseconds instead of seconds :) no need to think about parametrization.

src/app/shared/pipes/time-duration/timestamp.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { Timestamp } from './timestamp';
+
+describe('time-duration', () => {
```
dawidgil, line 3: Awesome, thanks! 👍

dawidgil (approved):
Looks good, just adjust this one test and i think you can merge it :) Thanks!

src/app/routes/enrolment/view-requests/components/expiration-info/expiration-info.component.html
whitneypurdum: sounds good!

PR #697, opened 2022-08-01
fix: show loader when getting list after notifications update

dawidgil:


nadia-adila (approved):

jrhender (approved):
Seems like solid changes, thanks @dawidgil

PR #699, opened 2022-08-01
fix(exp): do not send NaN as expiration date to ICL request

whitneypurdum:
- Was sending NaN to ICL as expirationTimestamp if date was removed. While this didn't affect ICL behavior as it is falsey, I think it should be changed to null in this case.

### DIFF
--- a/src/app/routes/enrolment/view-requests/components/expiration-date/expiration-date.component.spec.ts
+++ b/src/app/routes/enrolment/view-requests/components/expiration-date/expiration-date.component.spec.ts
@@ -40,6 +40,7 @@ describe('ExpirationDateComponent', () => {
     component = fixture.componentInstance;
     hostDebug = fixture.debugElement;
     spyOn(component.add, 'emit');
+    spyOn(component.remove, 'emit');
     jasmine.clock().install();
     baseTime = new Date(2000, 0, 1, 1, 0, 0);
     timezoneSeconds = baseTime.getTimezoneOffset() * 60;
@@ -90,7 +91,7 @@ describe('ExpirationDateComponent', () => {
     removeButton.click();
     fixture.detectChanges();
 
-    expect(component.remove.emit).toHaveBeenCalled;
+    expect(component.remove.emit).toHaveBeenCalled();
     expect(component.expirationDate.value).toEqual('');
   });
 
--- a/src/app/routes/enrolment/view-requests/components/expiration-date/expiration-date.component.spec.ts
+++ b/src/app/routes/enrolment/view-requests/components/expiration-date/expiration-date.component.spec.ts
@@ -82,7 +82,7 @@ describe('ExpirationDateComponent', () => {
     );
   });
 
-  it('should check if remove button sets expiration time as undefined', () => {
+  it('should check if remove button calls "remove" to set the expiration time as null', () => {
     component.defaultValidityPeriod = 100;
     fixture.detectChanges();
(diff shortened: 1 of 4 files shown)


src/app/routes/enrolment/view-requests/issuer-requests/issuer-requests.component.ts
```diff
@@ -83,8 +83,8 @@ export class IssuerRequestsComponent
     this.dialogRef.close(true);
   }
 
-  updateExpirationDate(validityPeriod: number): void {
-    this.expirationTime = Date.now() + validityPeriod;
+  updateExpirationDate(validityPeriod: number | undefined): void {
+    this.expirationTime = validityPeriod ? Date.now() + validityPeriod : null;
```
jrhender, line 87: IMO, this increases the complexity of the function a bit too much and it is better to avoid the use of `undefined`/`null` as a parameter if possible as it is hard to guess how the function will behave by reading its signature. In order words, by just reading `updateExpirationDate(validityPeriod: number | undefined): void`, it's hard to guess what the behaviour will be if `undefined` is passed. Does that mean that `expirationTime` will be set to `undefined` or does it mean that it will be set to some value and just `validityPeriod` won't be used?

I propose that there be two functions:
- `setExpirationTimeUsingValidity: void`
- `clearExpirationDate(): void`

In my opinion, this would make the code clearer.

whitneypurdum: Implemented this

src/app/routes/enrolment/view-requests/components/expiration-date/expiration-date.component.spec.ts
```diff
@@ -82,15 +82,16 @@ describe('ExpirationDateComponent', () => {
...
+  it('should check if remove button calls "remove" to set the expiration time as null', () => {
     component.defaultValidityPeriod = 100;
     fixture.detectChanges();
 
     const { removeButton } = getSelectors(hostDebug);
     removeButton.click();
     fixture.detectChanges();
 
-    expect(component.add.emit).toHaveBeenCalledWith(undefined);
+    expect(component.remove.emit).toHaveBeenCalled;
```
dawidgil, line 93: I think You need to add a spy for it like for `add`. And you lost brackets.
```suggestion
    expect(component.remove.emit).toHaveBeenCalled();
```

src/app/routes/enrolment/view-requests/components/expiration-date/expiration-date.component.ts
```diff
@@ -18,6 +18,7 @@ import { takeUntil } from 'rxjs/operators';
 export class ExpirationDateComponent implements OnInit, OnDestroy {
   @Input() defaultValidityPeriod: number;
   @Output() add: EventEmitter<number> = new EventEmitter<number>();
+  @Output() remove: EventEmitter<number> = new EventEmitter<number>();
```
dawidgil, line 21:
```suggestion
  @Output() remove: EventEmitter<void> = new EventEmitter<void>();
```
This way you won't accidentally emit value.

whitneypurdum: Ah good catch, I forgot to update! Thanks

dawidgil (approved):
Lgtm, thanks!

PR #700, opened 2022-08-01
fix: Do not show Cancel Claim Request if claim is Revoked

whitneypurdum:
- Do not show Cancel Claim Request button in "My Enrolments" view IF the enrolment is revoked on-chain, or if it is revoked off-chain. 

Screen Shot 2022-08-01 at 11 35 46 AM
Screen Shot 2022-08-01 at 11 35 31 AM

For some reason, I can;t override the min-height on the mat-menu-panel, even if I put !important. I wish I could because it looks a little too tall when there is just one button...

### DIFF
--- a/src/app/routes/enrolment/my-enrolment-list/my-enrolment-list.component.ts
+++ b/src/app/routes/enrolment/my-enrolment-list/my-enrolment-list.component.ts
@@ -93,6 +93,15 @@ export class MyEnrolmentListComponent implements OnInit, OnDestroy {
     return !element?.isSynced;
   }
 
+  showCancelClaimOption(element: EnrolmentClaim) {
+    return (
+      !element?.isAccepted &&
+      !element?.isRejected &&
+      !element.isRevokedOffChain &&
+      !element.isRevokedOnChain
+    );
+  }
+
   view(element: EnrolmentClaim) {
     this.dialog
       .open(ViewRequestsComponent, {
--- a/src/app/routes/enrolment/my-enrolment-list/my-enrolment-list.component.ts
+++ b/src/app/routes/enrolment/my-enrolment-list/my-enrolment-list.component.ts
@@ -93,15 +93,6 @@ export class MyEnrolmentListComponent implements OnInit, OnDestroy {
     return !element?.isSynced;
   }
 
-  showCancelClaimOption(element: EnrolmentClaim) {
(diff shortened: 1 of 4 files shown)


src/app/routes/enrolment/my-enrolment-list/my-enrolment-list.component.ts
```diff
@@ -93,6 +93,15 @@ export class MyEnrolmentListComponent implements OnInit, OnDestroy {
     return !element?.isSynced;
   }
 
+  showCancelClaimOption(element: EnrolmentClaim) {
+    return (
+      !element?.isAccepted &&
+      !element?.isRejected &&
+      !element.isRevokedOffChain &&
+      !element.isRevokedOnChain
```
jrhender, line 101: I would add the elvis (`?`) to avoid possible "cannot read property x of undefined" errors if these are reordered or the first couple are removed.
```suggestion
      !element?.isRevokedOffChain &&
      !element?.isRevokedOnChain
```

dawidgil: Why not to put this into EnrolmentClaim object? The reason i created this object was that we had a lot of copy pasted logic like this in components.
I mean, you could add method `canCancelClaim(): boolean`
Anyway isAccepted and isRejected is always available. No need to put elvis there.
Another thing it is good to add some tests.
@jrhender there will be no `"cannot read property x of undefined"` because the object (element) is available.

whitneypurdum: @dawidgil  It wasn't in the Enrolment claim object before today. I assumed that you had moved what you wanted into the object when you created it.

dawidgil: @whitneypurdum i might miss something always :D but do you agree that this might fits there or not really?

dawidgil (approved):
Looks good to me, big Thanks! :)

src/app/routes/enrolment/my-enrolment-list/my-enrolment-list.component.ts
whitneypurdum: @dawidgil Implemented the above in claim class

PR #701, opened 2022-08-02
fix: errors when reinitializing SB with wrong metamask network id

dawidgil:
https://energyweb.atlassian.net/browse/SWTCH-1379
Fix works like this:
1. When app detects metamask with wrong chainId but will try to reinitialize with wrong chainId, then popup with logout button will be displayed.
2. When metamask is not present, then it's doing nothing.
3. when chain id is wrong, but it will not try to reinitialize with metamask (for example user was using mobile connect), then nothing is doing.

### DIFF
--- a/src/app/shared/services/login/login.service.ts
+++ b/src/app/shared/services/login/login.service.ts
@@ -206,7 +206,7 @@ export class LoginService {
   public wrongNetwork() {
     if (this.isSessionActive()) {
       this.openSwal(
-        { title: 'Ops!', text: "You're using wrong network" },
+        { title: 'Oops!', text: "You're using the wrong network" },
         false
       );
     }
@@ -218,6 +218,7 @@ export class LoginService {
         icon: 'error',
         button: 'Proceed',
         closeOnClickOutside: false,
+        closeOnEsc: false,
         ...config,
       })
     )
@@ -235,7 +236,7 @@ export class LoginService {
       this.loadingService.hide();
       this.openSwal(
         {
-          title: 'Ops!',
+          title: 'Oops!',
           text: 'Something went wrong :(',
         },
         true
--- a/src/app/shared/services/login/login.service.ts
+++ b/src/app/shared/services/login/login.service.ts
@@ -11,8 +11,19 @@ import {
 import SWAL from 'sweetalert';
 import { from, Observable, of } from 'rxjs';
 import { IamListenerService } from '../iam-listener/iam-listener.service';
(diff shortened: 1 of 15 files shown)


src/app/shared/services/login/login.service.ts
```diff
@@ -199,6 +203,15 @@ export class LoginService {
     }
   }
 
+  public wrongNetwork() {
+    if (this.isSessionActive()) {
+      this.openSwal(
+        { title: 'Ops!', text: "You're using wrong network" },
```
whitneypurdum, line 209: "Oops!" (extra o). "You are using the wrong network" (add 'the')

whitneypurdum: Should we maybe specify "MetaMask network"?

src/app/state/auth/auth.selectors.spec.ts
```diff
@@ -24,7 +24,7 @@ describe('Auth Selectors', () => {
     it('should return false when metamaskChainId is set to Volta chain', () => {
       expect(
         authSelectors.isMetamaskDisabled.projector({
-          metamask: { chainId: '0x12047' },
+          metamask: { chainId: 73799 },
```
whitneypurdum, line 27: Do we want to move Ids to enum or variable or config file or something?

src/app/shared/services/login/login.service.ts
dawidgil: I can check for adding a button to import/switch to correct metamask configuration. But don't know if SWAL lib allows me to do it.

src/app/state/auth/auth.selectors.spec.ts
dawidgil: This might be a good idea to create something. I will think about it, what will be the best here :) 
Is there something you prefer?

whitneypurdum: No, whatever you think is best! A config file maybe that depends on env? As I think chain ID is dependent on that?

dawidgil: I created an object for storing EWC and Volta chain ids, reused in all places that i was able to find.

src/app/shared/services/login/login.service.ts
dawidgil: I was able to add it. However I will improve it by creating AlertService (wrapper for swal). This way will be easier to test functions that uses Swal. To be honest, i do not know how to mock swal 😅

whitneypurdum (approved):

PR #705, opened 2022-08-03
fix(exp): do not show 'Publish' button for expired claims

whitneypurdum:
- create getter (canPublishClaim) to determine if "Publish" button should be shown for pending claims on My Enrolments tab
- Add unit tests for isExpired and canPublishClaim getter


dawidgil (approved):
Looks good, thanks!

PR #706, opened 2022-08-03
fix(exp): use undefined vs null if no expirationTimestamp

whitneypurdum:
- ICL is expecting undefined or number for expirationTimestamp. In keeping with the expected types, just changing this to undefined from 'null' if no value. 
- As one step further, we could not add expirationTimestamp to the request object to ICL if timestamp is undefined...


jrhender (approved):
Seems okay to me 👍

dawidgil (approved):

PR #707, opened 2022-08-03
fix(exp): have validity period components accept miliseconds

whitneypurdum:
This addresses https://energyweb.atlassian.net/jira/software/projects/SWTCH/boards/54?selectedIssue=SWTCH-1389
- The fundamental issue was the validity period component was relying on seconds after the validity period (after clicking "Next" on new role form) was converted to milliseconds. As a result, if you input a validity period, clicked "next" and then clicked "back", the input fields did not have the correct numbers. 
- As a fix, I have this component now accepting milliseconds. There is no instance in any of the validity or expiration components where seconds are being used now. Everything is now using milliseconds only. I think this makes things easier as we don't have to divide by 1000 anywhere etc. 

I will test this more.

### DIFF
--- a/src/app/routes/applications/new-role/components/validity-period/validity-period.component.spec.ts
+++ b/src/app/routes/applications/new-role/components/validity-period/validity-period.component.spec.ts
@@ -44,7 +44,7 @@ describe('ValidityPeriodComponent', () => {
   });
 
   it('should update input with provided value', () => {
-    component.validityPeriod = 135;
+    component.validityPeriod = 120015;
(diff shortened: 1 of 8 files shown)


whitneypurdum:
@dawidgil would love to get your feedback on this. This removes any usage of seconds - only milliseconds are used for all units of time. I want to make sure you are okay with this as it updates a lot of the code you initially wrote.

dawidgil (approved):
It looks good, thanks for removing this code debt! 👍

PR #709, opened 2022-08-04
fix: creating role in domain that already exists

dawidgil:
The tests were not updated and they were passing, but they shouldn't.
Tests are updated + fixed implementation. Now it show error message that domain already exist

### DIFF
--- a/src/app/routes/applications/new-role/services/role-creation.service.ts
+++ b/src/app/routes/applications/new-role/services/role-creation.service.ts
@@ -21,10 +21,12 @@ export class RoleCreationService {
   async canUseDomain(domain: string): Promise<boolean> {
     try {
       const exists = await this.domainsFacade.checkExistenceOfDomain(domain);
-      const isOwner = await this.domainsFacade.isOwner(domain);
       if (exists) {
         return false;
       }
+
+      const isOwner = await this.domainsFacade.isOwner(domain);
+
       if (!exists && !isOwner) {
         return false;
       }
--- a/src/app/routes/applications/new-role/services/role-creation.service.ts
+++ b/src/app/routes/applications/new-role/services/role-creation.service.ts
@@ -21,10 +21,15 @@ export class RoleCreationService {
   async canUseDomain(domain: string): Promise<boolean> {
     try {
       const exists = await this.domainsFacade.checkExistenceOfDomain(domain);
-      if (!exists) {
-        return true;
+      const isOwner = await this.domainsFacade.isOwner(domain);
+      if (exists) {
+        return false;
       }
-      return await this.domainsFacade.isOwner(domain);
+      if (!exists && !isOwner) {
+        return false;
+      }
+
+      return !exists;
     } catch (e) {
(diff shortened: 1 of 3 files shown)


src/app/routes/applications/new-role/services/role-creation.service.ts
```diff
@@ -21,10 +21,15 @@ export class RoleCreationService {
   async canUseDomain(domain: string): Promise<boolean> {
     try {
       const exists = await this.domainsFacade.checkExistenceOfDomain(domain);
-      if (!exists) {
-        return true;
+      const isOwner = await this.domainsFacade.isOwner(domain);
+      if (exists) {
+        return false;
+      }
```
jrhender, line 27: Slightly more efficient to do it this way as we don't need `isOwner` if `exist == true`
```suggestion
      if (exists) {
        return false;
      }
      const isOwner = await this.domainsFacade.isOwner(domain);
```

src/app/routes/applications/new-role/services/role-creation.service.ts
```diff
@@ -21,10 +21,15 @@ export class RoleCreationService {
...
+      if (!exists && !isOwner) {
+        return false;
       }
-      return await this.domainsFacade.isOwner(domain);
+
+      return !exists;
```
jrhender, line 32: I'm confused, it seems to me like the effective implementation stayed the same 🤔 
Isn't the new logic effectively the same as the old logic?

src/app/routes/applications/new-role/services/role-creation.service.ts
dawidgil: You're right, I'll update this one.

src/app/routes/applications/new-role/services/role-creation.service.ts
dawidgil: nope, i updated the tests and they were failing. Previous set of tests and implementation was checking if domain exist, here method was saying canUseDomain.
So used functions are the same, but the logic and returned values are different.

jrhender: I don't understand how that's possible 🤔 .
My assumption is that `!exists` will always be `true` here because of this clause 
```
if (exists) {
  return false;
}
```
So we can substitute `!exists` for `true`. Then we have
```
if (true && !isOwner) {
        return false;
}
return true;
```
Then, by removing the redundant `true` and inverting the `if` statement we get
```
if (isOwner) {
 return true;
}
return false;
``` 
This can be simplified to `return isOwner`. If you remove the assignment to `isOwner`, then we have line currently on `develop`. What am I missing here?

dawidgil: Yeah, you're right. I removed most of the code, and tests are passing.

jrhender (approved):
This change seems good, but I'm not sure how it would fix the "role re-creation" bug. Was the bug due to the ownership check not being done correctly?

dawidgil:
Yes, it was returning true always, that's why async validation didn't worked correctly.

PR #711, opened 2022-08-04
fix(revocation): fix check for isRevoked logic

whitneypurdum:
- fix check to get into logic to check if a credential isRevoked

### DIFF
--- a/src/app/routes/enrolment/services/revoke/revoke.service.ts
+++ b/src/app/routes/enrolment/services/revoke/revoke.service.ts
@@ -50,7 +50,10 @@ export class RevokeService {
               'Claim Revoke'
             );
           } else {
-            this.toastrService.error('An error occured. On-chain claim was not revoked', 'Claim Revoke');
+            this.toastrService.error(
+              'An error occured. On-chain claim was not revoked',
+              'Claim Revoke'
+            );
           }
           return Boolean(value);
         }),
@@ -73,7 +76,10 @@ export class RevokeService {
               'Claim Revoke'
             );
           } else {
-            this.toastrService.error('An error occured. Off-chain claim was not revoked', 'Claim Revoke');
+            this.toastrService.error(
+              'An error occured. Off-chain claim was not revoked',
+              'Claim Revoke'
+            );
           }
           return Boolean(value);
         }),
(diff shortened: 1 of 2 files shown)


src/app/shared/services/claims-facade/claims-facade.service.ts
```diff
@@ -224,7 +224,7 @@ export class ClaimsFacadeService {
   }
 
   private async setIsRevokedOffChainStatus(enrolment: EnrolmentClaim) {
-    if (enrolment.isRevocableOffChain) {
```
whitneypurdum, line 227: claims that were revoked off chain already were never getting into this logic to check their revocation status (and therefore to set isRevokedOffChain), so it was always coming back as true and therefore the button was always 'active'. In this way, we are just checking each credential with a statusListEntry.

src/app/routes/enrolment/services/revoke/revoke.service.ts
```diff
@@ -43,7 +43,7 @@ export class RevokeService {
   private handleRevokeResponse() {
     return (source) =>
       source.pipe(
-        map((value: boolean) => {
+        map((value: boolean | StatusList2021Credential) => {
```
whitneypurdum, line 46: the revoke check for off-chain claims returns StatusList2021Credential, not boolean. No effect on anything with this change, but is more reflective of the truth.

whitneypurdum: Maybe we should have a different handler for each response? What is best practice for this @dawidgil ?

dawidgil: I think it's better to have different handler. Because when we came back to this in next few months we might not remembering that off-chain claims return StatusList.

dawidgil: However if you want to leave as it is for now, then i'm ok with it.

dawidgil (approved):

src/app/routes/enrolment/services/revoke/revoke.service.ts
whitneypurdum: No, I think separate is better too! Added here.

PR #712, opened 2022-08-05
Fix role name async validation

dawidgil:
Fixing again (😅) role name validation + small refactoring.
Issue was with wrong checked domain in method `isOwner`. And this lead me to create DomainUtils class for domain/namespace manipulation. I've changed only few places to reuse it.


jrhender (approved):

PR #714, opened 2022-08-10
Fix unable to complete login signature

dawidgil:
It fixes two issues:
https://energyweb.atlassian.net/browse/SWTCH-1399
https://energyweb.atlassian.net/browse/SWTCH-1397

The issue for 1399 was a race condition between getting Metamask details and reinitialization (reinitialization was first always). Now it tries to reinitialize user after getting metamask information. I've split reinitialization on two separate actions: one with metamask, second without.
Adds timer for reinitialization (1397) - 1 minute to reinitialize like while logging in. If user do nothing, it will display popup.

### DIFF
--- a/src/app/shared/services/login/login.service.ts
+++ b/src/app/shared/services/login/login.service.ts
@@ -3,7 +3,6 @@ import { IamService, PROVIDER_TYPE } from '../iam.service';
 import { LoadingService } from '../loading.service';
 import { ToastrService } from 'ngx-toastr';
 import {
-  AccountInfo,
   IS_ETH_SIGNER,
   ProviderType,
   PUBLIC_KEY,
@@ -107,7 +106,7 @@ export class LoginService {
   login(
     loginOptions?: LoginOptions,
     redirectOnChange = true
-  ): Observable<{ success: boolean; accountInfo?: AccountInfo | undefined }> {
+  ): Observable<{ success: boolean; }> {
     this.waitForSignature(redirectOnChange);
     return this.initialize(loginOptions, redirectOnChange);
   }
@@ -115,7 +114,7 @@ export class LoginService {
   reinitialize(
     loginOptions?: LoginOptions,
     redirectOnChange = true
-  ): Observable<{ success: boolean; accountInfo?: AccountInfo | undefined }> {
+  ): Observable<{ success: boolean; }> {
     this.loadingService.show();
     this.createTimer(1, redirectOnChange);
     return this.initialize(loginOptions, redirectOnChange);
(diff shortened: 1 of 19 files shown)


src/app/shared/services/login/login.service.ts
```diff
@@ -159,51 +173,25 @@ export class LoginService {
...
-    const messages = [
-      {
-        message: `Your ${connectionMessage}signature is being requested.`,
-        relevantProviders: 'all',
-      },
-      {
-        message:
-          'EW Key Manager should appear in a new browser tab or window. If you do not see it, please check your browser settings.',
-        relevantProviders: ProviderType.EwKeyManager,
-      },
```
jrhender, line 182: Some older stuff here. Good to clean this up, thanks 👍

src/app/state/auth/auth.effects.ts
```diff
@@ -186,39 +169,35 @@ export class AuthEffects {
...
+
+        return AuthActions.reinitializeAuth();
+      })
+    )
+  );
+
+  reinitializeLoggedUserWithMetamask$ = createEffect(() =>
+    this.actions$.pipe(
+      ofType(AuthActions.reinitializeAuthWithMetamask),
+      this.isMetamaskDisabled(),
```
jrhender, line 192: Is this line the key difference between `reinitializeLoggedUserWithMetamask` and `reinitializeLoggedUser`? Where in the first one, additional information about whether or not metamask is disabled is required?

dawidgil: Yes, isMetamaskDisabled means it has wrong chainId. I might rename it into for example `hasMetamaskCorrectChainId`.
`reinitializeLoggedUser` action is called when application tries reinitializing with different wallet than metamask (for example wallet connect).

dawidgil:
@jrhender i do not remember why we added `accountInfo` in returning from initializeConnection, however it is not used in the code as far as i see. I will remove it.

dawidgil:
Ok, i searched git history. accountInfo was added when we had staking inside switchboard (long time ago). We were displaying information like chain name and used wallet. This is not used to if you have nothing against i will remove it in this PR.

dawidgil:
@jrhender Ah, i badly understood you. Additional information from metamask is needed, because we do not know what chain id is used before trying to reinitialize. Chain id is used only for Metamask (wallet connect or ekc do not care about chainId).
I think we should not make any http call if user is using for example Ethereum network in MM and visiting Switchboard. By knowing this (if chainId === environment.chainId) I'm preventing calling login call to ssi-hub. Is it somehow explaining this?

whitneypurdum (approved):

PR #721, opened 2022-08-11
fix(infura): instantiate did registry with infura config

whitneypurdum:
- Sets Infura config values in .env files (maybe this should be in Github actions for privacy?)
- Instantiate connectToDidRegistry with Infura Auth

### DIFF
--- a/src/app/shared/services/iam.service.ts
+++ b/src/app/shared/services/iam.service.ts
@@ -187,7 +187,10 @@ export class IamService {
         this.cacheClient = cacheClient;
         this.verifiableCredentialsService = verifiableCredentialsService;
         if (createDocument) {
-          const { didRegistry, claimsService } = await connectToDidRegistry();
+          const { didRegistry, claimsService } = await connectToDidRegistry(
+             // @ts-ignore
+            this.configureIpfsConfig()
+          );
           this.didRegistry = didRegistry;
           this.claimsService = claimsService;
         }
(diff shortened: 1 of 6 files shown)


src/environments/environment.prod.ts
```diff
@@ -22,5 +22,7 @@ export const environment = {
   currencySymbol: 'EWT',
   blockExplorerUrl: 'https://explorer.energyweb.org',
   SENTRY_ENVIRONMENT: 'production',
+  INFURA_PROJECT_ID: '2DD5GcJRpMeuqMoKsgT95AQ8OI3',
```
whitneypurdum, line 25: I will update these values when we have dedicated projects

src/app/shared/services/iam.service.ts
```diff
@@ -267,6 +270,22 @@ export class IamService {
     return chainConfig;
   }
 
+  private configureIpfsConfig() {
```
whitneypurdum, line 273: I can write tests for this, if this is how we want it in this file.

src/app/shared/services/iam.service.ts
```diff
@@ -187,7 +187,10 @@ export class IamService {
         this.cacheClient = cacheClient;
         this.verifiableCredentialsService = verifiableCredentialsService;
         if (createDocument) {
-          const { didRegistry, claimsService } = await connectToDidRegistry();
+          const { didRegistry, claimsService } = await connectToDidRegistry(
+             // @ts-ignore
+            this.configureIpfsConfig()
```
whitneypurdum, line 192: We won't need this ignore once ICL pr is in and version is bumped, but maybe this would work now before it's in as John suggested.

src/environments/environment.prod.ts
dawidgil: If this is sensitive data, then we should put it into Github variables.

src/app/shared/services/iam.service.ts
dawidgil: I think tests are not necessary here. It's just config file.

dawidgil (approved):

PR #722, opened 2022-08-11
fix: disable enrolment button if user already requested enrolment

whitneypurdum:
Screen Shot 2022-08-11 at 5 13 44 PM
- Make call to get user requested claims. There was a placeholder there in the code to do so, but perhaps it was just forgotten. All variables were already in place, just no data fetched for them.

### DIFF
--- a/src/app/routes/search-result/search-result.component.ts
+++ b/src/app/routes/search-result/search-result.component.ts
@@ -2,7 +2,7 @@
 import { Component, OnInit, ViewChild } from '@angular/core';
 import { FormBuilder, FormControl, FormGroup } from '@angular/forms';
 import { ActivatedRoute } from '@angular/router';
-import { SearchType } from 'iam-client-lib';
+import { Claim, SearchType } from 'iam-client-lib';
 import { BehaviorSubject } from 'rxjs';
 import { debounceTime, distinctUntilChanged } from 'rxjs/operators';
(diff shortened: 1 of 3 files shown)


src/app/routes/search-result/search-result.component.ts
```diff
@@ -79,8 +79,11 @@ export class SearchResultComponent implements OnInit {
         this.filteredOptions.next(options);
       });
     this.activeRoute.queryParams.subscribe(async (queryParams: any) => {
-      // Get requested claims
       try {
+        this.requestedClaims =
```
dawidgil, line 83: 1. You can reuse data from store, no need to make another call. It's the same list as My enrolments.
2. Calling this method every time when query params changes isn't a good idea, because list of claims won't change.
My proposition is:
1. Inject store into constructor.
2. rename `requestedClaims` into `requestedClaims$`.
3. assign `requestedClaims$ = this.store.select(OwnedEnrolmentsSelectors.getEnrolments`
4. in html change `requestedClaims` into `requestedClaims$ | async`.
$ at the end of variable name informs that it is a stream.

whitneypurdum: thanks @dawidgil !

dawidgil (approved):
Looks good, thanks!

PR #726, opened 2022-08-12
fix: provider listeners callback

dawidgil:


src/app/shared/services/iam-listener/iam-listener.service.ts
```diff
@@ -42,12 +36,10 @@ export class IamListenerService {
     callback: (config) => void
   ): void {
     const { message, title } = this.getSwalConfigInfo(changeType);
-
     const config = {
       title,
       text: `${message} Please login again.`,
       icon: 'warning',
-      button: 'Proceed',
```
jrhender, line 50: The bug was this `button` property?

dawidgil: yes, if you open developer tools you will see the error. There are two options to set buttons and they are excluding each other. one with `button` second with `buttons`. I switched everything in login.service.ts into `buttons` (to add import config for wrong metamask configuration) but forgot about updating this part.
it is better imo to handle everything in one way instead of deleting `button` property to not get an error.

dawidgil: to summarize, i'm passing in callback `buttons` and here i was setting `button`. This was leading to an error.

jrhender (approved):

src/app/shared/services/iam-listener/iam-listener.service.ts
jrhender: Great, thanks for the explanation

PR #733, opened 2022-08-16
feat: show credential JSON on My Enrolments preview

whitneypurdum:
Feature to add legacy and new credential JSON to the View Enrolment
Changes (to 'My Enrolments' only): 
1. Change label on three dots from "View Request" to "View Credential" IF the claim has been issued or revoked
2. Change header on view pop up to 'Credential' if it has been issued or revoked, or 'Credential Request' if not yet approved
3. Show accordion for legacy eip171 if credential is issued
4. Show accordion for eip712 if credential is issued AND there is vp

https://www.loom.com/share/31088a799f3e40c89239cd0183d9da85

### DIFF
--- a/src/app/routes/enrolment/enrolment-list/components/actions/preview/preview.component.ts
+++ b/src/app/routes/enrolment/enrolment-list/components/actions/preview/preview.component.ts
@@ -21,7 +21,6 @@ export class PreviewComponent {
   private readonly VIEW_CREDENTIAL = 'View Credential';
   private readonly VIEW_REQUEST = 'View Request';
   @Input() enrolmentType: EnrolmentListType;
-  @Input() experimentalEnabled: boolean;
   @Input() set element(claim: EnrolmentClaim) {
     this.enrolmentClaim = claim;
     this.updateHeader(claim);
@@ -36,7 +35,6 @@ export class PreviewComponent {
       .open(GetPreviewComponent.get(this.enrolmentType), {
         width: '600px',
         data: {
-          experimentalEnabled: this.experimentalEnabled,
           claimData: this.enrolmentClaim,
         },
         maxWidth: '100%',
--- a/src/app/routes/enrolment/enrolment-list/components/actions/preview/preview.component.ts
+++ b/src/app/routes/enrolment/enrolment-list/components/actions/preview/preview.component.ts
@@ -18,17 +18,26 @@ import { GetPreviewComponent } from './get-preview-component';
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class PreviewComponent {
-  @Input() element: EnrolmentClaim;
+  private readonly VIEW_CREDENTIAL = 'View Credential';
+  private readonly VIEW_REQUEST = 'View Request';
   @Input() enrolmentType: EnrolmentListType;
+  @Input() experimentalEnabled: boolean;
+  @Input() set element(claim: EnrolmentClaim) {
+    this.enrolmentClaim = claim;
+    this.updateHeader(claim);
+  }
   @Output() refreshList = new EventEmitter();
+  header: string;
+  enrolmentClaim: EnrolmentClaim;
   constructor(private dialog: MatDialog) {}
 
-  view(element: EnrolmentClaim) {
+  view() {
     this.dialog
       .open(GetPreviewComponent.get(this.enrolmentType), {
         width: '600px',
         data: {
-          claimData: element,
+          experimentalEnabled: this.experimentalEnabled,
+          claimData: this.enrolmentClaim,
         },
         maxWidth: '100%',
         disableClose: true,
@@ -38,5 +47,12 @@ export class PreviewComponent {
       .subscribe(() => {
         this.refreshList.emit();
       });
+  }
+
+  updateHeader(claim: EnrolmentClaim) {
+    this.header =
+      claim.isAccepted || claim.isRevoked
+        ? this.VIEW_CREDENTIAL
+        : this.VIEW_REQUEST;
   }
 }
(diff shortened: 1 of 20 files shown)


jrhender:
Adding @JGiter and @nichonien as reviewers to give feedback while @dawidgil is away. This is a big enough feature that it probably makes sense to way for @dawidgil 's return before merging though, IMO

src/app/routes/enrolment/view-requests/components/full-credential/full-credential.component.ts
```diff
@@ -0,0 +1,13 @@
...
+
+@Component({
+  selector: 'app-credential-json',
+  templateUrl: './full-credential.component.html',
+  styleUrls: ['./full-credential.component.scss'],
+})
+export class CredentialJsonComponent {
+  @Input() title: string;
+  @Input() credentialData: object;
+  @Input() expEnabled: boolean;
```
jrhender, line 11: In my opinion, it would be better to use `experimentalEnabled` rather than `expEnabled`. Best to avoid abbreviations in names, IMO. Actually, when I first read this, I thought it might have been "expiryEnabled".

jrhender (commented):
Looks good from my perspective, just a small variable name comment

src/app/routes/enrolment/view-requests/components/full-credential/full-credential.component.ts
whitneypurdum: Resolving this as the flag has been removed.

src/app/routes/enrolment/models/enrolment-claim.ts
```diff
@@ -92,6 +94,10 @@ export class EnrolmentClaim
     );
   }
 
+  get verifiableCredential(): VerifiableCredential<ICredentialSubject> {
```
dawidgil, line 97: There is already this check in EnrolmentClaimAbstract under `credential`. Unfortunately interfaces are different, because it returns `VerifiableCredential<RoleCredentialSubject>`. However when I changed `credential` to return `VerifiableCredential<ICredentialSubject>` it works fine. So maybe we could merge those two getters?

src/app/shared/services/claims-facade/claims-facade.service.ts
```diff
@@ -176,6 +176,13 @@ export class ClaimsFacadeService {
               from(this.setIsRevokedOffChainStatus(enrolment))
             ),
           ])
+        ),
+        switchMap((enrolments: EnrolmentClaim[]) =>
+          forkJoin([
+            ...enrolments.map((enrolment) =>
+              from(this.setDecodedToken(enrolment))
```
dawidgil, line 183: Good idea to decode it when creating enrolment-claim object. 👍 We could decode both: token and issuedToken instead of doing it in view-requests component by using `TokenDecodeService`. But this doesn't need to be done now, we can refactor it next time :)

src/app/routes/enrolment/view-requests/view-requests.component.spec.ts
```diff
@@ -1,47 +1,120 @@
...
+        isIssued: false,
+        claimType: 'request-test',
+        verifiableCredential: {
+          id: '124',
+        },
+      },
+    });
+    fixture.detectChanges();
+    expect(component.header).toEqual('Credential Request');
+  });
```
dawidgil, line 119: Good set of tests! Thanks 👍

src/app/routes/enrolment/view-requests/view-requests.component.html
```diff
@@ -20,6 +20,24 @@ <h3 class="default-color mb-2">{{ claim?.roleName | uppercase }}</h3>
       </app-fields>
 
       <app-fields data-qa-id="issuer-fields" title="Issuer fields" [fieldsList]="issuerFields$ | async"> </app-fields>
+      <ng-container *ngIf="claim?.isIssued && claim?.verifiableCredential">
```
dawidgil, line 23: Personally i prefer to have all logic in typescript files, so you could create a getter:
```
get isRawVerifiableVisible(): boolean {
  return this.claim?.isIssued && this.claim?.verifiableCredential
}
```
However it comes from times when i did not test html in units. You already tested this logic so it's up to you if you want to move it into getters.

dawidgil (approved):
Looks good. 👍  I added few small comments about improvements for future :)

src/app/routes/enrolment/enrolment-list/components/actions/preview/preview.component.ts
```diff
@@ -18,17 +18,24 @@ import { GetPreviewComponent } from './get-preview-component';
...
-  @Input() element: EnrolmentClaim;
+  private readonly VIEW_CREDENTIAL = 'View Credential';
+  private readonly VIEW_REQUEST = 'View Request';
   @Input() enrolmentType: EnrolmentListType;
+  @Input() set element(claim: EnrolmentClaim) {
+    this.enrolmentClaim = claim;
+    this.updateHeader(claim);
+  }
   @Output() refreshList = new EventEmitter();
+  header: string;
```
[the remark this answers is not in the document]
whitneypurdum: oh awesome, yes.

src/app/routes/enrolment/models/enrolment-claim.ts
whitneypurdum: @dawidgil you are totally right, I got rid of this entirely and just used credential. Great.

src/app/routes/enrolment/view-requests/view-requests.component.html
whitneypurdum: @dawidgil I removed this and moved logic into enrolmentClaim (and added several tests)

PR #740, opened 2022-08-18
fix(assets): remove incorrect check for previously owned

whitneypurdum:
Screen Shot 2022-08-18 at 3 01 57 PM

### DIFF
--- a/src/app/routes/assets/asset-list/asset-list.component.html
+++ b/src/app/routes/assets/asset-list/asset-list.component.html
@@ -115,7 +115,7 @@
       </td>
     </ng-container>
 
-    <ng-container matColumnDef="owner" *ngIf="isPreviouslyOwned">
+    <ng-container matColumnDef="owner">
       <th mat-header-cell *matHeaderCellDef mat-sort-header>
         <ng-container>Offered By</ng-container>
       </th>


src/app/routes/assets/asset-list/asset-list.component.html
```diff
@@ -115,7 +115,7 @@
       </td>
     </ng-container>
 
-    <ng-container matColumnDef="owner" *ngIf="isPreviouslyOwned">
```
dawidgil, line 118: Yeah, it is wrong. But it should be `isOffered`.

whitneypurdum: Hmm I actually think it works without a check?

dawidgil: It works, but it is separation between those views in one component. Without that we do not know if this is for all tabs. later we might think if this is a bug, that this field is not displayed on the screen.

dawidgil: i would like in future to split this component into 3 components like in enrolments.

whitneypurdum: okay. There is already a check for the column types based on isOffered in the .ts file I think, which is why this column is not appearing for other asset views. But can add it.

dawidgil: if it is, then approved. Thanks for clarification and fixing.

dawidgil (approved):

PR #743, opened 2022-08-19
Hotfix/set revocation address

dawidgil:


whitneypurdum (approved):

src/app/shared/services/iam.service.ts
```diff
@@ -262,6 +262,8 @@ export class IamService {
   private getChainConfig(): Partial<ChainConfig> {
     const chainConfig: Partial<ChainConfig> = {
       rpcUrl: this.envService.rpcUrl,
+      claimsRevocationRegistryAddress:
+        '0xd72B4c8D5B1a1A4C7085259548bDF1A175CFc48D',
```
JGiter, line 266: where you get this address? In ICL it is different https://github.com/energywebfoundation/iam-client-lib/blob/018ad615c2556677a0451822327fc4466561190a/src/config/chain.config.ts#L65

dawidgil: from here https://energywebfoundation.slack.com/archives/C028ZC4G2RE/p1660825118078639?thread_ts=1660815524.703459&cid=C028ZC4G2RE

PR #752, opened 2022-08-30
fix: do not show enrolments for assets under 'My Enrolments'

whitneypurdum:
- When creating the list of enrolments under 'My Enrolments', filter out enrolment requests that you are NOT the subject of. Otherwise enrolments for your assets will show up under 'My Enrolments'.

### DIFF
--- a/src/app/state/enrolments/owned/owned.effects.ts
+++ b/src/app/state/enrolments/owned/owned.effects.ts
@@ -16,6 +16,11 @@ export class OwnedEnrolmentsEffects {
       tap(() => this.loadingService.show()),
       switchMap(() =>
         this.claimsFacade.getClaimsByRequester().pipe(
+          map((claims) => {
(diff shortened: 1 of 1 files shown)


whitneypurdum:
@jrhender Any thoughts on this PR? I think Dawid mentioned he thinks we should show asset enrolment requests under My Enrolements (conversation [here](https://energyweb.atlassian.net/jira/software/projects/SWTCH/boards/54?selectedIssue=SWTCH-1425)). At the very least, I think I should create a  PR that removes the button disable for enrolment if a DID is already enroled. The UI already handles if an asset is not enroled by a user is enroled (see screenshots on issue). The current UI is causing [this bug](https://energyweb.atlassian.net/jira/software/projects/SWTCH/boards/54?selectedIssue=SWTCH-1422).

jrhender:
@whitneypurdum I replied to the JIRA issue

PR #758, opened 2022-09-05
feat: use cascading filter in applications and roles views

dawidgil:


whitneypurdum:
Nice @dawidgil! No small undertaking. Was wondering if on “my Enrolments” and “enrolment requests” if we want to be able to filter by asset did as well? Not sure if this will work out of the box with the did filter, or if we even want to be able to filter by asset did…

whitneypurdum:
> Nice @dawidgil! No small undertaking. Was wondering if on “my Enrolments” and “enrolment requests” if we want to be able to filter by asset did as well? Not sure if this will work out of the box with the did filter, or if we even want to be able to filter by asset did…

We could just make another story for that too if we wanted to do it.

dawidgil:
@whitneypurdum it should work out of the box. Just change parameter in CascadingFilterComponent in `showDID = true`. Filter for DID is checking subject and requester (or requestor).

whitneypurdum (approved):

PR #762, opened 2022-09-06
fix (publish assets): pass subject DID to check for sync status

whitneypurdum:
Addresses bugs for Assets in 'My Enrolment Requests' view:
1. pass in subject DID (rather than signer DID) for sync checks (isSyncedOnChain, isSycnedOffChain) for your asset's enrolments. This fixes the issue of if you have an approved role for an asset, and you navigate to 'My Enrolment Requests' and you click "publish" for your asset's approved role, the publish is successful and the "publish" button disappears. (Previous behavior - when you click "publish", the button would never disappear as the sync checks for the enrolment were incorrect - they were using signer DID rather than asset did)
2. Add "Asset DID" column to "My Enrolments" view. This is to indicate that the enrolment is for your asset DID, not your DID. The table looks v. crowded with this in there.

### DIFF
--- a/src/app/shared/services/claims-facade/claims-facade.service.spec.ts
+++ b/src/app/shared/services/claims-facade/claims-facade.service.spec.ts
@@ -50,56 +50,24 @@ describe('ClaimsFacadeService', () => {
         claimType: createClaimType(claimType),
       } as Claim);
     };
-
-    it('should return empty list when getting empty list', async () => {
-      claimsServiceSpy.getUserClaims.and.returnValue(Promise.resolve([]));
-      expect(await service.addStatusIfIsSyncedOffChain([])).toEqual([]);
-    });
-
-    it('should return not changed list when UserClaims is empty', async () => {
-      claimsServiceSpy.getUserClaims.and.returnValue(Promise.resolve([]));
-      expect(
-        await service.addStatusIfIsSyncedOffChain([createClaim('123')])
-      ).toEqual([
-        jasmine.objectContaining({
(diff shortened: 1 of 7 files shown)


src/app/shared/services/claims-facade/claims-facade.service.spec.ts
```diff
@@ -50,56 +50,24 @@ describe('ClaimsFacadeService', () => {
         claimType: createClaimType(claimType),
       } as Claim);
     };
-
-    it('should return empty list when getting empty list', async () => {
-      claimsServiceSpy.getUserClaims.and.returnValue(Promise.resolve([]));
-      expect(await service.addStatusIfIsSyncedOffChain([])).toEqual([]);
-    });
-
-    it('should return not changed list when UserClaims is empty', async () => {
```
dawidgil, line 59: Those tests are removed because they are no longer valid or they are testing nothing in your opinion?

dawidgil (approved):
Approved, just one comment about unit tests :)

src/app/shared/services/claims-facade/claims-facade.service.spec.ts
whitneypurdum: Oh I just changed them actually because it’s not a list anymore, it’s mapping over each claim rather than the list. It’s still testing the map though I think…

dawidgil: Yeah, right! Thanks for clarify, you can merge it imo :)

PR #765, opened 2022-09-06
fix: add check for undefined

whitneypurdum:
- addresses this error: https://sentry.io/organizations/energy-web/issues/3544473359/?project=5651521&query=is%3Aunresolved&statsPeriod=14d

### DIFF
--- a/src/app/modules/cascading-filter/services/cascading-filter/cascading-filter.service.ts
+++ b/src/app/modules/cascading-filter/services/cascading-filter/cascading-filter.service.ts
@@ -115,7 +115,7 @@ export class CascadingFilterService {
           .status(this.statusFilter)
           .organization(this.organizationFilter)
           .build()
-          .map((claim) => claim.organization)
+          .map((claim) => claim?.organization)
       )
     );
   }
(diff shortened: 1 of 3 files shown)


dawidgil:
I don't think here is the issue. On the list of enrolments there is an undefined instead of EnrolmentClaim. Probably that is causing issues with undefined. In fact every enrolment claim should contain organization, because every role belongs to organization (and might belong to application). So this might remove the errors, but not the issue that is causing it.
Does it (⬆️ ) make sense to you?

src/app/modules/cascading-filter/services/cascading-filter/cascading-filter.service.ts
```diff
@@ -115,7 +115,7 @@ export class CascadingFilterService {
           .status(this.statusFilter)
           .organization(this.organizationFilter)
           .build()
-          .map((claim) => claim.organization)
+          .map((claim) => claim?.organization)
```
jrhender, line 118: Is it okay to have `undefined` values in the `map` array?

src/app/state/enrolments/owned/owned.selectors.ts
```diff
@@ -21,6 +21,6 @@ export const getEnrolments = createSelector(
 export const getNotSyncedAmount = createSelector(
   getAllEnrolments,
   (enrolments) => {
-    return enrolments.filter((enrolment) => enrolment.canPublishClaim).length;
+    return enrolments?.filter((enrolment) => enrolment.canPublishClaim).length;
```
jrhender, line 24: Is it okay to potentially have `undefined` passed as a second parameter to `createSelector`?

PR #771, opened 2022-09-13
fix(DIDBook): update clear button functionality

whitneypurdum:
DID Book form:
- Add a separate button for "Clear" (so that Clear and Cancel functionality belong to separate components)
- Only show 'Cancel' button if it is for single record (i.e. Not DID book)
- Only show 'Clear' button if it is for DID book
- Disable "Clear" unless there is text in the DID or Label field
- Small CSS changes

With https://www.loom.com/share/2a92888a2dd048b5a2aadab1b86d3bca

For bug: https://energyweb.atlassian.net/jira/software/projects/SWTCH/boards/54?selectedIssue=SWTCH-1427


src/app/modules/did-book/components/did-book-form/did-book-form.component.html
```diff
@@ -1,30 +1,35 @@
...
-  <div class="d-flex justify-content-end mb-2">
-    <button mat-stroked-button class="btn-small mr-2" data-qa-id="cancel" (click)="reject()">Cancel</button>
+  <div class="d-flex justify-content-between mb-2">
+    <button
+      mat-raised-button
+      class="btn btn-small btn btn-primary"
+      data-qa-id="cancel"
+      (click)="reject()"
+      [disabled]="clearDisabled">
+      Clear
```
dawidgil, line 40: This can't be hardcoded clear button, because this component can be "closed" in different view. Anyway it emits cancel event, which is misleading when you change it's name to clear.

whitneypurdum: Thanks @dawidgil , I responded in the issue - thanks for commenting there and clarifying this.

whitneypurdum: @dawidgil I separated the functionality of clear and cancel into 2 buttons. Thanks!

src/app/modules/did-book/components/did-book-form/did-book-form.component.ts
```diff
@@ -19,6 +19,7 @@ import { DidBookRecord } from '../models/did-book-record';
 export class DidBookFormComponent implements OnInit {
   @Input() did: string;
   @Input() label: string;
+  @Input() isDidBook: boolean;
```
dawidgil: Maybe more about the context inside this component, not outside? isDidBook tells nothing about this component. 
I prefer to add here 2 inputs like:
```suggestion
  @Input() showClear: boolean;
  @Input() showCancel: boolean;
```
This tells more about behaviour in this component. What do you think?

whitneypurdum: Thanks @dawidgil, made these changes

dawidgil (approved):

PR #773, opened 2022-09-15
task/SWTCH-1530 update iam-client-lib@7.0.0-alpha.1

JGiter:


package.json
```diff
@@ -49,7 +49,7 @@
     "@ew-did-registry/credentials-interface": "^0.6.3-alpha.596.0",
     "@ew-did-registry/did-resolver-interface": "^0.6.3-alpha.596.0",
     "@fortawesome/fontawesome-free": "5.5.0",
-    "@gnosis.pm/safe-apps-sdk": "^7.3.0",
+    "@gnosis.pm/safe-apps-provider": "^0.13.1",
```
jrhender: Could you explain why the change from `sdk` to `provider` again please?

JGiter: It turned out that there is no need to install `safe-apps-sdk`, because it is dependency of `safe-apps-provider`

JGiter: And despite that switchboard-dapp uses only `safe-apps-sdk` I would like to ensure that its version is the same that in iam-client-lib, but iam-client-lib depends on `safe-apps-provider`

jrhender: But why install `provider` in SB if only `sdk` is needed? Is `provider` also needed?

jrhender: > And despite that switchboard-dapp uses only safe-apps-sdk I would like to ensure that its version is the same that in iam-client-lib, but iam-client-lib depends on safe-apps-provider

Ok, I mean if we are bitting the bullet of nested dependency imports, can we just rely on the installation of `safe-app-provider` (and `sdk`) from ICL then?

JGiter: We used extraneous dependencies at the begining, but later decided to install dependencies explicitly used by switchboard. I don't remember exact reason though. May be that was because when there are several packages depends on `safe-apps-sdk` then it is not clear which one will be hoisted to the root and used by switchboard. May be its not so bad as none other of switchboard depencies uses gnosis

jrhender: Any thoughts on this @dawidgil and @whitneypurdum ?

dawidgil: Imo if it is possible, i would remove this dependency from SB and make ICL install required dependency. Afaik SB do not use it directly (we have only config for it). Everything is done on ICL side, so imo SB might know nothing about it.

JGiter: Alright. Lets do this way

JGiter: By the way, following this approach we can also uninstall ew-did-registry dependencies

dawidgil: @whitneypurdum correct me if I'm wrong, but as far as i remember, you installed them, because some interfaces were not exported by ICL, is that right? If yes, then we should first reexport them and then remove it.

whitneypurdum: @dawidgil  IServiceEndpoint, VerifiableCredential, VpRequestInteractService, IDIDDocument are all imported from ew-did-reg, but they were there before I started developing in this code base. Maybe they are all exported now from ICL, so if so, we can replace and remove.

dawidgil: Cool, thanks for clarify :) So @JGiter if you want to remove them, you need to check if required interfaces are reexported from ICL and refactor it :) for me, you have green light to do that :D if you do not want to do this, just leave it and we will refactor it

JGiter: Lets remove them in separate PR

PR #782, opened 2022-09-30
fix: displaying expiry date when issuing a vc

dawidgil:
https://energyweb.atlassian.net/browse/SWTCH-1566

### DIFF
--- a/src/app/shared/expiration/date-picker/expiration-date.component.ts
+++ b/src/app/shared/expiration/date-picker/expiration-date.component.ts
@@ -19,7 +19,16 @@ export class ExpirationDateComponent implements OnInit, OnDestroy {
   /**
    * Defined in milliseconds.
    */
-  @Input() defaultValidityPeriod: number;
+  @Input() set defaultValidityPeriod(period: number) {
+    if (!period) {
+      return;
+    }
+    this._defaultPeriod = period;
+    this.defaultExpirationDate();
+  }
+  get defaultValidityPeriod() {
+    return this._defaultPeriod;
+  }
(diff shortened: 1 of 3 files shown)


whitneypurdum (approved):

src/app/shared/expiration/date-picker/expiration-date.component.ts
```diff
@@ -19,14 +19,24 @@ export class ExpirationDateComponent implements OnInit, OnDestroy {
   /**
    * Defined in milliseconds.
    */
-  @Input() defaultValidityPeriod: number;
+  @Input() set defaultValidityPeriod(period: number) {
+    if (!period) {
+      return;
+    }
+    this._defaultPeriod = period;
+    this.defaultExpirationDate();
```
jrhender, line 27: Maybe this is a side point, but I find understanding the difference between `defaultExpirationDate` and `setDefaultExpirationDate` challenging. It seems like `setDefaultExpirationDate` has the same code as `defaultExpirationDate`, maybe `setDefaultExpirationDate` could call `defaultExpirationDate` and the function names could be adjusted accordingly?

dawidgil: ahh, yeah. Code looks similar. I will adjust it, thanks for pointing this out.

dawidgil: https://github.com/energywebfoundation/switchboard-dapp/pull/783 hope this is what you wanted :)

jrhender: Thanks!

PR #783, opened 2022-09-30
refactor: remove code duplication in expiration date

dawidgil:


src/app/shared/expiration/date-picker/expiration-date.component.ts
```diff
@@ -78,15 +78,12 @@ export class ExpirationDateComponent implements OnInit, OnDestroy {
...
     this.add.emit(this.defaultValidityPeriod);
-    this.expirationDate.setValue(
-      new Date(Date.now() + this.defaultValidityPeriod)
-    );
-    this.expirationTimeShift = this.defaultValidityPeriod;
     this.hideRemoveButton = false;
   }
 
-  defaultExpirationDate() {
+  defaultExpirationDate(): void {
```
jrhender: I'm curious what your personal convention is for using the `set` prefix? Why (or why not) should this have `set` prefix (i.e. be `setDefaultExpirationDate()`). It seems as though `this.expirationDate` is being set so maybe it makes sense?

src/app/shared/expiration/date-picker/expiration-date.component.ts
```diff
@@ -78,15 +78,12 @@ export class ExpirationDateComponent implements OnInit, OnDestroy {
   }
 
   restoreDefaultHandler(): void {
+    this.defaultExpirationDate();
     this.add.emit(this.defaultValidityPeriod);
-    this.expirationDate.setValue(
-      new Date(Date.now() + this.defaultValidityPeriod)
-    );
-    this.expirationTimeShift = this.defaultValidityPeriod;
```
jrhender, line 85: This line doesn't seem to be in the new logic. Is this expected?

src/app/shared/expiration/date-picker/expiration-date.component.ts
```diff
@@ -78,15 +78,12 @@ export class ExpirationDateComponent implements OnInit, OnDestroy {
   }
 
   restoreDefaultHandler(): void {
+    this.defaultExpirationDate();
     this.add.emit(this.defaultValidityPeriod);
```
jrhender: This is a bit odd to me. `expirationDate` is being set but we're emitting `defaultValidityPeriod`. But maybe I'm missing some pattern/convention here. Does it make sense to you?

jrhender (commented):
Thanks for trying out the refactor! Left a few questions

src/app/shared/expiration/date-picker/expiration-date.component.ts
dawidgil: Yeah, defaultExpirationDate was setting expirationDate to be equal default, so it worked, but logically it wasn't correct. Changed it by adding method:
```
  private emitExpirationTimeShift() {
    this.add.emit(this.expirationTimeShift);
  }
```

src/app/shared/expiration/date-picker/expiration-date.component.ts
dawidgil: It is, in method `setDefaultExpirationDate` (changed it from `defaultExpirationDate`).

src/app/shared/expiration/date-picker/expiration-date.component.ts
dawidgil: added prefix, you're right. I'm setting there a value, so it's better to add `set`.

jrhender (approved):
Looks good, thanks!

PR #784, opened 2022-10-05
Extension for e2e

dawidgil:
Required changes to have working E2E tests for checking organisation and creating roles

### DIFF
--- a/src/app/routes/applications/new-role/new-role.component.ts
+++ b/src/app/routes/applications/new-role/new-role.component.ts
@@ -73,7 +73,6 @@ export class NewRoleComponent implements OnInit, AfterViewInit {
     }
   }
 
-  IssuerTypes = [IssuerType.DID, IssuerType.ROLE];
   revoker: IRoleType;
   issuer: IRoleType;
   signerDID = this.signerFacade.getDid();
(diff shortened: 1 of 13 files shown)


src/app/routes/applications/new-role/new-role.component.ts
```diff
@@ -618,6 +617,10 @@ export class NewRoleComponent implements OnInit, AfterViewInit {
     this.goNextStep();
   }
 
+  isLastStep() {
+    return this.stepper?.steps?.length - 1 === this.stepper?.selectedIndex;
```
jrhender, line 621: Would we want `isLastStep` to return `true` if `this.stepper` is undefined? I think this is what we would get here as `undefined === undefined` is `true`, I believe. So maybe worth checking if `this.stepper` is undefined?
```suggestion
    return this.stepper && this.stepper?.steps?.length - 1 === this.stepper?.selectedIndex;
```

dawidgil: It might be when initializing component/view. That's why i added elvis operator. It should work without it as well.

jrhender (approved):
Looks good!

PR #786, opened 2022-10-06
Remove entirely infura secrets

dawidgil:


.env.example
```diff
@@ -0,0 +1,3 @@
+#Example of .env file. Fill below secrets for local development.
+#INFURA_PROJECT_ID=
+#INFURA_PROJECT_SECRET=
```
jrhender, line 3: Could you update the README with instructions of what one should do with `.env.example` in order to run Switchboard locally please?

src/app/shared/services/env/env.service.ts
```diff
@@ -21,7 +21,7 @@ export class EnvService {
   blockExplorerUrl: string = environment.blockExplorerUrl;
   SENTRY_ENVIRONMENT: string = environment.SENTRY_ENVIRONMENT;
   INFURA_PROJECT_ID: string =
-    INFURA_PROJECT_ID ?? environment?.INFURA_PROJECT_ID;
+    INFURA_PROJECT_ID ?? process?.env?.INFURA_PROJECT_ID;
   INFURA_PROJECT_SECRET: string =
-    INFURA_PROJECT_SECRET ?? environment?.INFURA_PROJECT_SECRET;
+    INFURA_PROJECT_SECRET ?? process?.env?.INFURA_PROJECT_SECRET;
```
jrhender, line 26: Do we also need to change the GHA jobs or will the secrets also be available from `process.env` (it seems like they were previously available via `environment`)?

jrhender (commented):
Added a couple questions

src/app/shared/services/env/env.service.ts
dawidgil: No need to change GHA jobs. This process.env is available locally and it is reading .env file. On GHA there is no file like this, so this part is not changed at all.

.env.example
dawidgil: Should i add link to the confluence page? Or just information that user needs to provide those values?

jrhender: Hmm, good question. I think the instructions can assume that the person has Infura credentials. I think we need some steps like:

1. Execute `cp .env.example .env`
2. Fill in necessary variables

Does that make sense from your perspective?

dawidgil: Yup, added to the readme. Hope the place where i added it is good in you opinion.

README.md
```diff
@@ -31,6 +31,12 @@ Install dependencies:
 npm install
 ```
 
+### Running
+
+Before you start You need to provide some secrets:
+1. Execute cp .env.example .env
```
jrhender:
```suggestion
1. Execute `cp .env.example .env`
```

.env.example
jrhender: Yes, looks good, thanks!

jrhender (approved):
Minor suggestion to add backtick to instruction, but looks good, thanks!

PR #793, opened 2022-10-18
fix: did contact errors - setting validators and getting list.

dawidgil:


src/app/modules/did-book/services/did-book.service.ts
```diff
@@ -29,7 +30,11 @@ export class DidBookService {
   getList(): void {
     this.httpDidBook
       .getList()
-      .subscribe((list: DidBookRecord[]) => this.list.next(list));
+      .pipe(retryWhenWithDelay())
```
jrhender, line 33: Does `retryWhen` typically have an argument which defines when to retry? This is what I'm understanding from here: https://www.learnrxjs.io/learn-rxjs/operators/error_handling/retrywhen

If so, why no argument in this case?

dawidgil: This is not retryWhen from rxjs. It is retryWhenWithDelay and is defined [here](https://github.com/energywebfoundation/switchboard-dapp/blob/develop/src/app/operators/retry-when-with-delay/retry-when-with-delay.ts)

jrhender: Ah gotcha, thanks for the explanation!

src/app/modules/did-book/components/did-book-form/did-book-form.component.ts
```diff
@@ -25,7 +25,13 @@ export class DidBookFormComponent implements OnInit {
...
       return;
     }
-    this.form.get('did').addValidators(ListValidator.stringExist(dids));
+    this.form
+      .get('did')
+      .setValidators([
+        Validators.required,
+        HexValidators.isDidValid(),
+        ListValidator.stringExist(dids),
+      ]);
```
jrhender, line 34: Was this the root cause of the error then? The validators weren't correctly being set? Or rather that something a missing or invalid DID was passed in?

jrhender (approved):
Seems reasonable!

src/app/modules/did-book/components/did-book-form/did-book-form.component.ts
dawidgil: Validation was wrong. Steps to repro:
1. Add new DID contact
2. Remove this DID contact
3. try to add the same contact as in 1 (same DID). -> error show up, that this DID is already used. which is wrong obviously. The validators weren't cleaning up, this fixed the issue with cleaning up validators.

dawidgil: But this fixes different issue. I hope that retryWhenWithDelay will fix it. If not, then i will try to find another solution (i couldn't reproduce it locally).

PR #798, opened 2022-10-20
fix: add Switchboard label to Jira ticket and fix title

whitneypurdum:
- Updates Jira ticket title
- Adds 'Switchboard' label to Jira ticket

Addresses these comments: https://energyweb.atlassian.net/browse/SWTCH-1624

### DIFF
--- a/.github/workflows/create-issue.yml
+++ b/.github/workflows/create-issue.yml
@@ -24,11 +24,24 @@ jobs:
         with:
           project: SWTCH
           issuetype: Bug
-          summary: 'GitHub Issue ${{ github.event.issue.number }}: ${{ github.event.issue.title }}'
+          summary: '${{ github.event.issue.title }} (GitHub Issue ${{ github.event.issue.number }})'
           description: '${{ github.event.issue.body }} See issue: ${{ github.event.issue.html_url }}'
 
       - name: Save Issue ID
-        id: save-id
+        id: save-issue-id
         run: |
           ID='${{ steps.create.outputs.issue }}'
           echo ::set-output name=issueId::${ID}
+
+  add-switchboard-label:
+    runs-on: ubuntu-latest
+    needs: save-issue-id
+    steps:
+      - name: Login to Jira
+        uses: atlassian/gajira-login@v2.0.0
+        env:
+          JIRA_BASE_URL: ${{ secrets.JIRA_BASE_URL }}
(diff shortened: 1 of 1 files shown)


.github/workflows/create-issue.yml
```diff
@@ -24,11 +24,24 @@ jobs:
...
+    needs: save-issue-id
+    steps:
+      - name: Login to Jira
+        uses: atlassian/gajira-login@v2.0.0
+        env:
+          JIRA_BASE_URL: ${{ secrets.JIRA_BASE_URL }}
+          JIRA_USER_EMAIL: ${{ secrets.JIRA_USER_EMAIL }}
+          JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
+      - name: Add label
+        run: jira labels add ${{ needs.save-issue-id.outputs.issueId }} 'Switchboard'
```
jrhender: @whitneypurdum how does this work without a `uses`? For comparison, in the `Create JIRA Ticket` step, we have `uses: atlassian/gajira-create@v2.0.1`.

Maybe another way of asking the question is: where is the `jira labels add` function defined?

whitneypurdum: @jrhender John I’m not sure. I looked at many examples on GitHub and saw this command in multiple repos. If you aren’t comfortable merging it I can remove.

jrhender: I wouldn't say that we need to remove necessarily. Would just be good to understand it a bit more. Do you have any links to the examples that you could share please?

whitneypurdum: Sure @jrhender I saw it in these 2 repos:
https://github.com/dbt-labs/jira-actions/blob/322893b6b2c91c064e49a312ae91c1502babed22/.github/workflows/jira-label.yml#L62

https://github.com/dbt-labs/jira-actions/blob/c02f7b7553dc7841cea7bfa96753fc8485926b84/.github/workflows/jira-creation.yml#L108

jrhender: Thanks. In both examples, I see there is a set to setup the Jira cli:
```
      - name: Setup Jira
        uses: atlassian/gajira-cli@v2.0.2
```
I would that would be needed in our case as well. Do you think it needs to be added to this PR?

whitneypurdum: @jrhender right, I forgot to add. It is there now, thanks.

jrhender (approved):
Looks good, thanks!

PR #799, opened 2022-10-21
fix: displaying revoked enrolments while filtering by status

dawidgil:


jrhender:
@dawidgil is this related to a Jira task? Or just a bug you noticed?

dawidgil:
bug that i noticed. In table there was information `Revoked` but enrolments like this was displayed when you select Approved status in filter.

whitneypurdum (approved):

PR #800, opened 2022-10-26
chore: loggers for iam initialisation

dawidgil:
Loggers to check what's going on with not initialised DomainsService from ICL.


jrhender (commented):
Looks good to me. But let's plan to merge tomorrow morning after webinar

PR #801, opened 2022-10-26
fix: not disabled retry button when counting down

dawidgil:
Bug that i noticed after removing error with double running change detection

### DIFF
--- a/src/app/routes/applications/new-application/new-application.component.html
+++ b/src/app/routes/applications/new-application/new-application.component.html
@@ -40,16 +40,18 @@
         <ng-container *ngFor="let step of txs; let i = index">
           <mat-step [editable]="false" [completed]="false">
             <ng-template matStepLabel>{{ step.info }}</ng-template>
-            <app-card-info [description]="step.info"></app-card-info>
-            <div
-              class="card card-hint d-flex flex-row justify-content-start align-items-center shadow-none d-flex pt-3 pb-2 mb-3">
-              <mat-icon class="px-2">notifications_none</mat-icon>
-              <span class="pl-3 pb-2"> Please confirm this transaction in your connected Wallet. </span>
-            </div>
-            <div class="d-flex justify-content-end mb-3">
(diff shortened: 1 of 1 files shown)


src/app/routes/applications/new-application/new-application.component.html
```diff
@@ -40,16 +40,18 @@ <h4 mat-dialog-title class="mb-0 d-flex justify-content-between align-items-cent
...
-            <div
-              class="card card-hint d-flex flex-row justify-content-start align-items-center shadow-none d-flex pt-3 pb-2 mb-3">
-              <mat-icon class="px-2">notifications_none</mat-icon>
-              <span class="pl-3 pb-2"> Please confirm this transaction in your connected Wallet. </span>
-            </div>
-            <div class="d-flex justify-content-end mb-3">
-              <button mat-stroked-button class="btn btn-small mr-2" (click)="closeDialog()">Cancel</button>
-              <button mat-raised-button class="btn btn-small btn-primary" (appRetryBtn)="retry()">Retry</button>
-            </div>
+            <ng-container *ngIf="stepper?.selectedIndex === i + 2">
```
jrhender, line 43: @dawidgil I don't quite get it. Why the `2` here?

dawidgil: We have 2 steps before ^^ First for describing values for application, second for summary. That's why there is 2. It's far from ideal, but i do not want to refactor to much, so i'm doing my best to adjust it.

jrhender:
@dawidgil could you describe the repro steps for this bug please? I think that might help me understand the code change

dawidgil:
Sure.
1. Go to governance -> create application.
2. Fill data for new application
3. Go through details forward and you will see retry button.
4. Accept metamask request signature.
5. You will go to the next step. 
Issue: The count down in retry button do not reset, so instead counting down from 60, it's continuing counting from previous step (for example you will have 30seconds). The farther you'll go, the less you will have in the clock.

whitneypurdum (approved):

PR #812, opened 2022-11-01
Update jira gh action

whitneypurdum:
- add env variables to add label step

### DIFF
--- a/.github/workflows/create-issue.yml
+++ b/.github/workflows/create-issue.yml
@@ -34,18 +34,14 @@ jobs:
   add-switchboard-label:
     runs-on: ubuntu-latest
     needs: create-issue
+    env:
+      JIRA_BASE_URL: ${{ secrets.JIRA_BASE_URL }}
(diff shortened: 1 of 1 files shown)


.github/workflows/create-issue.yml
```diff
@@ -45,3 +45,7 @@ jobs:
           JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
       - name: Add label
         run: jira labels add ${{ needs.create-issue.outputs.issueId }} 'switchboard'
+        env:
+          JIRA_BASE_URL: ${{ secrets.JIRA_BASE_URL }}
+          JIRA_USER_EMAIL: ${{ secrets.JIRA_USER_EMAIL }}
+          JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
```
jrhender: I think these can be added in a shared way instead of being duplicated. This is how it was in one of the examples that you shared previously with me, as I recall

nichonien (approved):

jrhender (approved):

PR #813, opened 2022-11-01
feat: create mail form to request org in PRD

whitneypurdum:
This PR:
IF environment.production, then:
 1. the "Create Organization" button reads "Request to Create Organization"
 2. On click, the "Create Organization" button launches a mail to:
 
Screen Shot 2022-11-01 at 12 47 04 PM
NOTE that the current mailTo email is JH's email address (set as constant), but we will change when we have an IAM address.

View on Loom: https://www.loom.com/share/75af637d3a3449fc9e76f5c1ceeb662a

### DIFF
--- a/src/app/routes/applications/applications.component.spec.ts
+++ b/src/app/routes/applications/applications.component.spec.ts
@@ -1,24 +1,78 @@
 import { ComponentFixture, TestBed } from '@angular/core/testing';
-
+import { MatDialog } from '@angular/material/dialog';
+import { ActivatedRoute, Router } from '@angular/router';
+import { DialogService } from 'src/app/shared/services/dialog/dialog.service';
+import { EnvService } from 'src/app/shared/services/env/env.service';
+import { IamService } from 'src/app/shared/services/iam.service';
+import { UrlParamService } from 'src/app/shared/services/url-param.service';
+import { provideMockStore } from '@ngrx/store/testing';
+import { MockActivatedRoute, dialogSpy } from '@tests';
 import { ApplicationsComponent } from './applications.component';
+import { DomainsService } from 'iam-client-lib';
 
-xdescribe('ApplicationsComponent', () => {
+describe('ApplicationsComponent', () => {
   let component: ApplicationsComponent;
   let fixture: ComponentFixture<ApplicationsComponent>;
-
-  beforeEach(() => {
-    TestBed.configureTestingModule({
-      declarations: [ApplicationsComponent],
-    }).compileComponents();
+  let dialog;
+  let domainsService;
+  const routerSpy = jasmine.createSpyObj('Router', [
+    'navigateByUrl',
+    'navigate',
+  ]);
+  let activatedRouteStub: MockActivatedRoute;
+  describe('NON-PRODUCTION BEHAVIOR', () => {
+    beforeEach(() => {
+      dialog = jasmine.createSpyObj(DialogService, ['open']);
+      activatedRouteStub = new MockActivatedRoute();
+      domainsService = jasmine.createSpyObj('domainsService', ['isOwner']);
+      TestBed.configureTestingModule({
+        declarations: [ApplicationsComponent],
+        providers: [
+          { provide: MatDialog, useValue: dialogSpy },
+          { provide: EnvService, useValue: { production: false } },
+          { provide: IamService, useValue: { domainsService: domainsService } },
+          { provide: UrlParamService, useValue: {} },
+          { provide: Router, useValue: routerSpy },
+          { provide: ActivatedRoute, useValue: activatedRouteStub },
+          provideMockStore(),
+        ],
+      }).compileComponents();
+      fixture = TestBed.createComponent(ApplicationsComponent);
+      component = fixture.componentInstance;
+      fixture.detectChanges();
+      domainsService.isOwner.and.returnValue(Promise.resolve(true));
+    });
+    it('should create Request Organization Button text correctly based on env (non-production)', () => {
+      expect(component.orgRequestButtonText).toBe('Create Organization');
+    });
   });
 
-  beforeEach(() => {
-    fixture = TestBed.createComponent(ApplicationsComponent);
-    component = fixture.componentInstance;
-    fixture.detectChanges();
-  });
-
-  it('should create', () => {
-    expect(component).toBeTruthy();
+  describe('PRODUCTION BEHAVIOR', () => {
+    beforeEach(() => {
+      dialog = jasmine.createSpyObj(DialogService, ['open']);
+      activatedRouteStub = new MockActivatedRoute();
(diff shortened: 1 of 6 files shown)


src/app/routes/applications/applications.component.spec.ts
```diff
@@ -1,24 +1,78 @@
...
-    fixture = TestBed.createComponent(ApplicationsComponent);
-    component = fixture.componentInstance;
-    fixture.detectChanges();
+  let dialog;
+  let domainsService;
+  const routerSpy = jasmine.createSpyObj('Router', [
+    'navigateByUrl',
+    'navigate',
+  ]);
+  let activatedRouteStub: MockActivatedRoute;
```
whitneypurdum, line 22: @dawidgil would love your thoughts on this approach. Maybe this is overkill to have tests at all for this, but I am trying to practice my Angular testing :) 
I am doing a before each before each test suite here because I need to override the ENS Service provider, and one solution I read about was to just compile it for each suite, because if you call compileComponents() at the top, the values cannot be overridden per describe.

src/app/shared/services/env/env.service.ts
```diff
@@ -13,7 +13,7 @@ export class EnvService {
   showAzureLoginOption: boolean = environment.showAzureLoginOption;
   natsEnvironmentName: string = environment.natsEnvironmentName;
   rootNamespace: string = environment.rootNamespace;
-
+  production: boolean = environment.production;
```
dawidgil, line 16: This is only for prod build. It doesn't mean production (EWT chain), because `environment.stg.ts` also have `production: true`, but it deploys on Volta network.

src/environments/constants.ts
```diff
@@ -1,3 +1,4 @@
 export const constants = {
   trxRetry: 60, // 1 minute
+  orgRequestEmail: 'john.henderson@energyweb.org', //to do: create generic IAM email for inbound requests
```
dawidgil, line 3: Email will be only for EWT network, is that right? So maybe moving this to environment.prod.ts file would be better, and in constants make it as an empty string? Or maybe we should move it to gh variables? what do you think?

src/app/routes/applications/applications.component.ts
```diff
@@ -91,34 +94,37 @@ export class ApplicationsComponent implements OnInit, AfterViewInit, OnDestroy {
...
-      maxWidth: '100%',
-      disableClose: true,
-    });
-
-    dialogRef
-      .afterClosed()
-      .pipe(takeUntil(this.subscription$))
-      .subscribe((result) => {
-        // console.log('The dialog was closed');
+  handleNewOrgRequest(): void {
```
dawidgil, line 95:
```suggestion
  if (this.envService.production) {
    this.requestOrgCreationFromEmail()
   } else {
     if(!this.isIamEwcOwner) {
       this.requestOrgCreationFromEnrolment();
       return;
     } 
     this.createNewOrg();
   }
```
What do you think about split this method into few smaller?

src/app/routes/applications/applications.component.spec.ts
dawidgil: No, it's a good idea to have tests where we have some logic.
Yeah, they cannot be overridden, but you can mock EnvService (and override it in test suites), but this need some adjustment in code. 
Example how you can mock it:
```
describe('ApplicationsComponent', () => {
  let component: ApplicationsComponent;
  let fixture: ComponentFixture<ApplicationsComponent>;
  let dialog;
  let domainsService;
  const routerSpy = jasmine.createSpyObj('Router', [
    'navigateByUrl',
    'navigate',
  ]);
  let envService = {production: false};
  let activatedRouteStub: MockActivatedRoute;
  beforeEach(() => {
    dialog = jasmine.createSpyObj(DialogService, ['open']);
    activatedRouteStub = new MockActivatedRoute();
    domainsService = jasmine.createSpyObj('domainsService', ['isOwner']);

    TestBed.configureTestingModule({
      declarations: [ApplicationsComponent],
      providers: [
        { provide: MatDialog, useValue: dialogSpy },
        { provide: EnvService, useValue: envService },
        { provide: IamService, useValue: { domainsService: domainsService } },
        { provide: UrlParamService, useValue: {} },
        { provide: Router, useValue: routerSpy },
        { provide: ActivatedRoute, useValue: activatedRouteStub },
        provideMockStore(),
      ],
    }).compileComponents();
    fixture = TestBed.createComponent(ApplicationsComponent);
    component = fixture.componentInstance;
    domainsService.isOwner.and.returnValue(Promise.resolve(true));
  });
  it('should create Request Organization Button text correctly based on env (non-production)', () => {
    envService.production = false;
    console.log(envService);
    fixture.detectChanges();
    expect(component.orgRequestButtonText).toBe('Create Organization');
  });

  it('should create Request Organization Button text correctly based on env (production)', () => {
    envService.production = true;
    fixture.detectChanges();
    expect(component.orgRequestButtonText).toBe(
      'Request to Create Organization'
    );
  });
});
```
and in applications.component.ts move assigning
```  
orgRequestButtonText = this.envService.production
    ? 'Request to Create Organization'
    : 'Create Organization';
``` 
to ngOnInit.

Also if you do not want to mock so many things, you can always create new component and move only connected functionality there. Anyway it's fine if you leave it here :)

src/app/shared/services/env/env.service.ts
whitneypurdum: Ohh @dawidgil good call. Do we need another way to target production (EWT chain), or think its okay to have this in staging as well?

src/environments/constants.ts
whitneypurdum: @dawidgil yes, I will move it to prod.ts file then

src/app/shared/services/env/env.service.ts
whitneypurdum: Actually maybe this is okay because in stg you want to test production behavior anyways, so would be good to test this in stg before it goes to prd.

src/app/routes/applications/applications.component.ts
whitneypurdum: I split this into 2 methods

src/app/shared/services/env/env.service.ts
dawidgil: We can check the chain for example. Or check depending on `orgRequestEmail` property if it exists. If yes, then use email, if it is empty string, then org creator should work. It's up to you :)

whitneypurdum: @dawidgil I think this approach as we have is good because we will want to make sure in staging that this is working as expected based on production build.

whitneypurdum:
@jrhender or @mhrsntrk should the email not also include the requester's DID? How will you know who to give the org name ownership to otherwise?

dawidgil (approved):

PR #745, opened 2022-08-19
Revert "Hotfix/set revocation address"

dawidgil:
Reverts energywebfoundation/switchboard-dapp#743

PR for reverting hotfix with revocation address. When ICL with configured revocation address will be ready to merge, this PR should be merged first.

PR #817, opened 2022-11-07
1.3 release to prod

jrhender:


jrhender:
ICL in this release is at ["7.0.0-alpha.3"](https://github.com/energywebfoundation/switchboard-dapp/blob/61ab007bcba4891f8a006c837d73c5e3b33d54a4/package.json#L68) would be high enough to remove the revocation registry address hotfix I think. https://github.com/energywebfoundation/switchboard-dapp/pull/745

@dawidgil would we want to merge #745 before or after this PR. I think we would want to merge after so that there is no downtime, right? Or it is necessary to merge it before?

dawidgil:
I think after is a better idea, because we would have always valid code.

dawidgil (approved):

PR #819, opened 2022-11-07
fix: add var for orgRequestEmail in all envs

whitneypurdum:
- Add a var for orgRequestEmail for each env. and remove from constants

### DIFF
--- a/src/environments/environment.dev.ts
+++ b/src/environments/environment.dev.ts
@@ -8,7 +8,7 @@ export const environment = {
 
   rpcUrl: 'https://volta-rpc-vkn5r5zx4ke71f9hcu0c.energyweb.org/',
   chainId: ChainId.Volta,
-  cacheServerUrl: 'https://identitycache-dev.energyweb.org/v1',
+  cacheServerUrl: 'http://localhost:3000/v1',
   natsServerUrl: 'https://identityevents-dev.energyweb.org/',
   ekcUrl: 'https://azure-proxy-server.energyweb.org/api/v1',
   showAzureLoginOption: true,
@@ -23,5 +23,6 @@ export const environment = {
   SENTRY_ENVIRONMENT: 'develop',
   idleTime: 900,
   idleTimeout: 300,
+  orgRequestEmail: '',
   ...constants,
 };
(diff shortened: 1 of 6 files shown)


src/environments/environment.dev.ts
```diff
@@ -8,7 +8,7 @@ export const environment = {
 
   rpcUrl: 'https://volta-rpc-vkn5r5zx4ke71f9hcu0c.energyweb.org/',
   chainId: ChainId.Volta,
-  cacheServerUrl: 'https://identitycache-dev.energyweb.org/v1',
+  cacheServerUrl: 'http://localhost:3000/v1',
```
dawidgil: hmm, i think you need to revert this one line :)

src/environments/environment.stg.ts
```diff
@@ -23,5 +23,6 @@ export const environment = {
   SENTRY_ENVIRONMENT: 'staging',
   idleTime: 900,
   idleTimeout: 300,
+  orgRequestEmail: 'john.henderson@energyweb.org',
```
dawidgil: add this line to the environment.prod.ts (for release branch), as this stg is for staging (for master branch).

src/environments/environment.dev.ts
jrhender: This seems wrong or at least unrelated?

src/environments/environment.stg.ts
whitneypurdum: @dawidgil its already in prd

jrhender: I think devops will make the email soon, so I'm betting that this will avoid needing to change it down the road
```suggestion
  orgRequestEmail: 'iamteam@energyweb.org@energyweb.org',
```

jrhender (changes requested):
Need to check cacheserverurl

src/environments/environment.dev.ts
whitneypurdum: oh I don't know how I pushed this!

src/environments/environment.stg.ts
dawidgil: if you prefer we can always move it to gh variables :)

dawidgil (approved):

jrhender (approved):

PR #831, opened 2023-01-17
feat(sentry): add performance tracing

dawidgil:
Adds performance tracing to the sentry. Also i updated sentry to the latest version.


src/app/shared/services/sentry/sentry.service.ts
```diff
@@ -12,16 +12,17 @@ export class SentryService {
     return Sentry.init({
       dsn: SENTRY_DSN,
       environment: this.envService.SENTRY_ENVIRONMENT,
+      integrations: [new BrowserTracing()],
       tracesSampleRate: 0.2,
     });
   }
 
   error(message: string) {
-    Sentry.captureMessage(message, Severity.Error);
```
jrhender, line 20: Is `Severity` no longer included in `@sentry/angular`?

dawidgil: It's deprecated.There is info in interface:
`Please use a SeverityLevel string instead of the Severity enum. Acceptable values are 'fatal', 'error', 'warning', 'log', 'info', and 'debug'.`

dawidgil: where:
`export declare type SeverityLevel = 'fatal' | 'error' | 'warning' | 'log' | 'info' | 'debug';`

jrhender: Gotcha, thanks

jrhender (approved):

src/app/shared/services/sentry/sentry.service.ts
dawidgil: I had to update one more place with this SeverityLevel, because application wasn't building.

PR #837, opened 2023-02-09
chore: setup switchboard deployment for gp4btc-staging

Timtech4u:
Once merged to **develop**, this would have to be promoted to **gp4btc-staging**

### DIFF
--- a/.github/workflows/cf-deploy.yml
+++ b/.github/workflows/cf-deploy.yml
@@ -2,7 +2,7 @@ name: Deploy to bucket
 
 on:
   push:
-    branches: ['develop', 'master', 'sandbox', 'release']
+    branches: ['develop', 'master', 'sandbox', 'release', 'DOI-2408-gp4btc-staging', 'gp4btc-staging']
 jobs:
   cancel-previous:
     name: 'Cancel Previous Runs'
@@ -28,6 +28,9 @@ jobs:
           elif [ "${{ github.ref }}" = "refs/heads/master" ]; then
             echo "::set-output name=env_name::staging"
             echo "::set-output name=build_script::npm run build-staging"
+          elif [ "${{ github.ref }}" = "refs/heads/DOI-2408-gp4btc-staging" ]; then
+            echo "::set-output name=env_name::staging"
+            echo "::set-output name=build_script::npm run build-staging"            
           elif [ "${{ github.ref }}" = "refs/heads/release" ]; then
             echo "::set-output name=env_name::prod"
             echo "::set-output name=build_script::npm run build-prod"
--- a/.github/workflows/cf-deploy.yml
+++ b/.github/workflows/cf-deploy.yml
@@ -2,7 +2,7 @@ name: Deploy to bucket
 
 on:
   push:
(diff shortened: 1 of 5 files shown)


.github/workflows/cf-deploy.yml
```diff
@@ -28,6 +28,9 @@ jobs:
           elif [ "${{ github.ref }}" = "refs/heads/master" ]; then
             echo "::set-output name=env_name::staging"
             echo "::set-output name=build_script::npm run build-staging"
+          elif [ "${{ github.ref }}" = "refs/heads/DOI-2408-gp4btc-staging" ]; then
```
dawidgil: @Timtech4u Would it be possible to remove handling deployments from repository? I would like to just deploy docker images and that's all. Everything else (where it will be deployed) will be handled by you.

```suggestion
          elif [ "${{ github.ref }}" = "refs/heads/gp4btc-staging" ]; then
```

Timtech4u: @dawidgil  are you referring to the outdated commit or generally? More context would be helpful

dawidgil: No, I wrote nothing about commit. I'm talking about deployment switchboard.
What would happen if we will have 4 more instances of switchboard in different places? You will add more branches to handle that and they should be deployed under different circumstances?
for example: develop, master, staging, staging-2, staging-3, ..., staging-x, release ?
I would like to have develop, master and release. Nothing more. 
To make that working i could just send somewhere docker images and you automatically via k8s deploy to proper instances.

Timtech4u: @dawidgil  I think it's fine to refactor or close this PR based on the approach discussed from Slack. cc: @jrhender

dawidgil (approved):
Approved. It looks like we're going for now with multiple branches approach.

.github/workflows/cf-deploy.yml
```diff
@@ -28,6 +28,9 @@ jobs:
           elif [ "${{ github.ref }}" = "refs/heads/master" ]; then
             echo "::set-output name=env_name::staging"
             echo "::set-output name=build_script::npm run build-staging"
+          elif [ "${{ github.ref }}" = "refs/heads/gp4btc-staging" ]; then
+            echo "::set-output name=env_name::staging"
+            echo "::set-output name=build_script::npm run build-staging"
```
jrhender: I think we need a new `npm` command to execute the build with the `gp4btc-staging` values: https://energyweb.atlassian.net/browse/DOI-2417
```suggestion
            echo "::set-output name=build_script::npm run build-gp4btc-staging"            
```

jrhender (changes requested):

.github/workflows/cf-deploy.yml
dawidgil: @jrhender Will we have different backend for that staging env?

dawidgil: Are we planning to have more instances for gp4btc? For example dev/staging/prod? If not i will follow with what we already have in repo, so the script will be named `build-gp4btc`

dawidgil:
@jrhender scripts for building gp4btc-prod/stg are created.

jrhender (approved):

PR #842, opened 2023-02-16
Improve enrolments performance

dawidgil:
Part 1 of the improvement for: https://energyweb.atlassian.net/browse/SWTCH-1811
- Reuses fetched enrolments after login: works for my enrolments and issuer enrolments.
- Stop getting list of enrolments while changing tabs
- While navigating to enrolment view, it's getting only once revokable enrolments.
- Updating like rejecting / approving should fetch all claims from backend, filter them and get new version of claim with selected id, then it should update state. 
- removing claim should just remove from state that enrolment without additional requests (from my enrolments and issuer enrolments because i can enrol for role that i'm issuer).

In next PR i will try to improve nats updates.


jrhender (approved):
I did some quick testing locally, seems great 🚀 

One question about the PR description (which is great btw, thank you)
> Updating like rejecting / approving should fetch all claims from backend, filter them and get new version of claim with selected id, then it should update state. 

After rejecting/approving, why do all the all claims need to be fetched from backend. Can't just the single claim that was rejected/approved be fetched?

dawidgil:
hmm, maybe you're right. Maybe i don't need to fetch all of them, i will check ICL and swagger for improvement here! Thanks :)

dawidgil:
@jrhender yes, there is a method `getClaimById`. I will refactor it in part 2 of those improvements.

PR #843, opened 2023-02-16
Improve nats updates

dawidgil:
Part 2 of: https://github.com/energywebfoundation/switchboard-dapp/pull/842
- update specific enrolment in my list and issuer list depending on nats notification.

@jrhender you wanted to split those PRs, however this one must be merged after part 1 as this depends on code created there.

### DIFF
--- a/src/app/routes/enrolment/enrolment.component.html
+++ b/src/app/routes/enrolment/enrolment.component.html
@@ -32,7 +32,7 @@
           <app-my-enrolment-list
             [list]="myEnrolmentList$ | async"
             (refreshList)="refreshMyEnrolmentsList($event)"
-            (removeEnrolment)="removeEnrolmentFromMyList($event)">
+            (removeEnrolment)="removeEnrolment($event)">
           </app-my-enrolment-list>
         </div>
       </div>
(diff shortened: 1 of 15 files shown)


dawidgil:
Found out, that i can't use `getClaimsById` because i can't assume that user is requestor and issuer. So, when getting nats notification i will get information that "something has changes" with given claimId. I need to find it where and modify state with it.
There is still a place for improvement, because enrolment.component.ts is dispatching action for updates, also nats notifications are dispatching actions for updates. I could "stop" one action if similar action was dispatched in for example last 2-3 seconds.

jrhender (approved):
Looks good to me, thanks

jrhender:
> Found out, that i can't use getClaimsById because i can't assume that user is requestor and issuer. 

I think you can't assume that the user is both requestor **and** issuer, but I think you can assume that the user is authorized to get the claim. Is this not the case?

dawidgil:
~~No, not really.~~ Suppose I'm requesting to claim A. I'm not the issuer of this role. If i use getClaimById method i will get the claim, but i can't add that claim to the issuer list. `getClaimById` do not distinguish what "claim type" i want.

edit: wait, above example is true. Yes i can get that claim, but by getting it i have no idea what to do with it. By getting claims by issuer/revoker/requestor i have this separation.

dawidgil:
To fix above, i would need 3 new endpoints, getClaimBy(Issuer/Requestor/Revoker)Id.

jrhender:
> edit: wait, above example is true. Yes i can get that claim, but by getting it i have no idea what to do with it. By getting claims by issuer/revoker/requestor i have this separation.

It could be possible to figure it out by checking the `requester` field of the claim against the user or by comparing against the list of roles that the user is allowed to issue. But I understand that this is a complicated situation that could have edge cases, so I think it's okay, we can leave it for now

PR #845, opened 2023-02-21
fix(issuer-requests): throwing errors when opening view

dawidgil:
fixes https://energyweb.atlassian.net/browse/SWTCH-2042

### DIFF
--- a/src/app/routes/enrolment/view-requests/issuer-requests/issuer-requests.component.ts
+++ b/src/app/routes/enrolment/view-requests/issuer-requests/issuer-requests.component.ts
@@ -19,7 +19,7 @@ export class IssuerRequestsComponent
   extends ViewRequestsComponent
   implements OnInit
 {
-  @ViewChild('issuerFields', { static: false }) requiredFields: EnrolmentForm;
+  @ViewChild('issuerFields', { static: true }) requiredFields: EnrolmentForm;
   roleDefinition: IRoleDefinitionV2;
   expirationTime: number;
   requestorFields: Record<string, string | number> = {};
--- a/src/app/routes/registration/enrolment-form/enrolment-form.component.spec.ts
+++ b/src/app/routes/registration/enrolment-form/enrolment-form.component.spec.ts
@@ -371,6 +371,7 @@ describe('EnrolmentFormComponent', () => {
     fixture.detectChanges();
     expect(updateFormSpy).toHaveBeenCalled();
   });
+
   it('should update the appropriate form field with the copied input provided', () => {
     component.fieldList = [
       {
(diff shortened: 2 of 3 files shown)


src/app/routes/registration/enrolment-form/enrolment-form.component.spec.ts
```diff
@@ -387,6 +388,40 @@ describe('EnrolmentFormComponent', () => {
...
+        required: true,
+      },
+    ];
+
+    component.toCopy = { 'label 2': 'The label' };
+
+    fixture.detectChanges();
+    const formValue = (component.enrolmentForm?.get('fields') as FormArray)
+      ?.controls[0];
+    expect(formValue.value).toBeFalsy();
```
jrhender, line 406: Is the `label 1` field falsy because it should never have been set and the default is a falsy value (whereas `'The label'` is truthy)?

src/app/routes/enrolment/view-requests/issuer-requests/issuer-requests.component.ts
```diff
@@ -19,7 +19,7 @@ export class IssuerRequestsComponent
   extends ViewRequestsComponent
   implements OnInit
 {
-  @ViewChild('issuerFields', { static: false }) requiredFields: EnrolmentForm;
+  @ViewChild('issuerFields', { static: true }) requiredFields: EnrolmentForm;
```
jrhender, line 22: Was this change part of the fix? I'm reading this blog post to understand this better
https://angular.io/guide/static-query-migration

So is this change necessary because the query results need to be available in `ngOnInit`?

jrhender (commented):
Seems fine, added a couple more comments

src/app/routes/registration/enrolment-form/enrolment-form.component.spec.ts
dawidgil: Algorithm is comparing labels. The problem is with setting value. We're passing at start empty object, getting keys (empty list), and first element. Later we're trying to find `undefined` in another list. Algorithm treat is as it exists, but it might not exist.
The test is checking what will happen if the label that i want to "copy" doesn't exist in fields.

src/app/routes/enrolment/view-requests/issuer-requests/issuer-requests.component.ts
dawidgil: It solved issue with `ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked`. If i enable prod build for dev env, then it will disappear in all places. However it solved issue here for non prod. `issuerFields` are not under `ngIf` so they can be static. If this would be parametrised, so issuerFields are toggled via ngIf, then it should be `static: false`.

jrhender (approved):
Thanks for the additional explanations

PR #846, opened 2023-02-22
fix: error with closeConnection method call when iam service isn't initialized

dawidgil:
fix for: https://energyweb.atlassian.net/browse/SWTCH-1878


jrhender (approved):
👍

PR #848, opened 2023-02-23
fix: close connection error, as signerService might not be initialised

dawidgil:
Previous fix was wrong, as we need to have signer service initialised. IamService is always available => method exist.


jrhender (approved):

jrhender:
@dawidgil does it make sense to merge this?

dawidgil:
@jrhender yes, for me it's a fix for `closeConnection` error in sentry.

PR #849, opened 2023-02-24
fix: error with getting allowed roles by issuer

dawidgil:
Should fix this error: https://energyweb.atlassian.net/browse/SWTCH-1756
It might call `getAllowedRolesByIssuer` method from domainsService, while domainsService wasn't yet initialised.

### DIFF
--- a/src/app/modules/issue-vc/services/issuance-vc.service.ts
+++ b/src/app/modules/issue-vc/services/issuance-vc.service.ts
@@ -2,9 +2,12 @@
 import { Injectable } from '@angular/core';
 import { IamService } from '../../../shared/services/iam.service';
 import { LoadingService } from '../../../shared/services/loading.service';
-import { finalize, map, tap } from 'rxjs/operators';
+import { finalize, map, switchMap, tap } from 'rxjs/operators';
 import { Claim, IRole, RegistrationTypes } from 'iam-client-lib';
 import { Observable } from 'rxjs';
+import { Store } from '@ngrx/store';
+import { AuthSelectors } from '@state';
+import { truthy } from '@operators';
 
 @Injectable({
   providedIn: 'root',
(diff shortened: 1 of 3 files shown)


src/app/modules/issue-vc/services/issuance-vc.service.ts
```diff
@@ -2,9 +2,12 @@
 import { Injectable } from '@angular/core';
 import { IamService } from '../../../shared/services/iam.service';
 import { LoadingService } from '../../../shared/services/loading.service';
-import { finalize, map, tap } from 'rxjs/operators';
+import { finalize, map, switchMap, tap } from 'rxjs/operators';
```
jrhender, line 5: why the addition of `switchMap` here? it doesn't seem to be used
```suggestion
import { finalize, map, tap } from 'rxjs/operators';
```

jrhender (commented):
Your PR description mentions `domain.service` but this PR seems to mostly related to the `issuance-vc.service.ts`. Is the issuance service called from the domain service?

src/app/modules/issue-vc/services/issuance-vc.service.ts
dawidgil: switchMap probably left after testing. I need to improve eslint checking.

Yeah, issuance-vc service is using domain.service. And it wasn't waiting for user login. If user is not logged in, then there is no domain service. However `issuance-vc service` exists and is calling method when component was created. We had race condition here. I removed it, by getting list of roles from domain.service after successful login.

Imo, sometimes i think we should change login behaviour in SB to be similar to gp4btc. App is automatically navigating to login page and is trying to login if session is active. We can add easily feature with returnUrl like in SB. This means every call in app is done after successful login. 
In Sb we are displaying for example assets page and calling all methods while app setup for that page, but user might not be logged in yet. This means we need to check every method from ICL if we can use it.
But this is bigger change in app behaviour so i did not do that.

jrhender (approved):
Ok, sounds good, thanks for the explanations

PR #853, opened 2023-03-08
feat(iam.service): set domain to window origin

jrhender:
Setting SIWE domain to have appropriate signature message.
Similar to https://github.com/energywebfoundation/gp4btc/pull/257

### DIFF
--- a/src/app/shared/services/iam.service.ts
+++ b/src/app/shared/services/iam.service.ts
@@ -66,6 +66,9 @@ export class IamService {
     // Set Cache Server
     setCacheConfig(envService.chainId, {
       url: envService.cacheServerUrl,
+      auth: {
+        domain: window.location.origin,
+      },
     });
 
     // Set RPC


dawidgil (approved):
Just a question, cannot it be handled by ICL internally?

JGiter (approved):

jrhender:
> Just a question, cannot it be handled by ICL internally?

Yes, maybe we should improve this to avoid other apps needing similar changes. Or maybe we should make it required to set it and throw error if not set. A consideration is what we should set it to server-side if the `window` is not available.
What do you think @JGiter ?

JGiter:
>  A consideration is what we should set it to server-side if the `window` is not available

@jrhender Probably we can omit `domain` in Node.js client. In my opinion this property is used to communicate to user that some service wants user to authenticate to it, but in case of Node.js there is no external service

PR #863, opened 2023-03-17
Send source maps to sentry

dawidgil:
feat for: https://energyweb.atlassian.net/browse/SWTCH-2194
Motivation: It is hard to detect what part of code is failing in Sentry, because code is minified and uglified. Adding source maps will fix it. 
Adding source maps won't affect security as switchboard code is open source project.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -225,7 +225,7 @@
               "ngswConfigPath": "ngsw-config.json",
               "localize": ["std"]
             },
-            "dev": {
+            "develop": {
               "fileReplacements": [
                 {
                   "replace": "src/environments/environment.ts",
--- a/angular.json
+++ b/angular.json
@@ -225,7 +225,7 @@
               "ngswConfigPath": "ngsw-config.json",
               "localize": ["std"]
             },
-            "development": {
+            "dev": {
               "fileReplacements": [
                 {
                   "replace": "src/environments/environment.ts",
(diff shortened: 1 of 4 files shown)


angular.json
```diff
@@ -252,6 +252,14 @@
               ],
               "serviceWorker": true,
               "ngswConfigPath": "ngsw-config.json"
+            },
+            "development": {
```
jrhender, line 256: I think it's quite confusing to have a `dev` and a `development` environment. Is one of them for local development? If so, could we use `local`?

dawidgil: Yeah, you're right. It should be local. I will change that

dawidgil: I see that nx is creating project with `development` and `develop` as a build for dev env. Maybe that would be a better option?

jrhender: Sorry, I'm not following your comment here. Are you saying that nx doing both of these things:
1. creating project with `development`
2. creating `develop` as a build for dev env

dawidgil: nx creates project with `develop` and `development` as configuration option. See this:
image
First one is for building for dev environment (minifying, uglifying, tree shaking etc.) like for prod, second is for starting application locally.

jrhender: Ok, gotcha. Thanks for the screenshot

jrhender (approved):

PR #864, opened 2023-03-17
chore(deps): update angular and related packages to v14

dawidgil:
pr for task: https://energyweb.atlassian.net/browse/SWTCH-2196


dawidgil:
@jrhender there are few `as any` because Angular v14 supports typed form. However we have some forms created on the fly, which creates issues with typing. If you prefer i can add unit tests + refactor/rewrite those forms to have proper typing. Do you want from me to rewrite in this PR or should I leave it as it is?

angular.json
```diff
@@ -226,32 +226,12 @@
...
-                  "with": "src/environments/environment.dev.ts"
-                }
-              ],
-              "optimization": true,
-              "outputHashing": "all",
-              "sourceMap": true,
-              "namedChunks": false,
-              "extractLicenses": true,
-              "vendorChunk": false,
-              "buildOptimizer": true,
```
jrhender: Is this removal intentional? Is it related to angular 14 upgrade?

src/app/routes/applications/new-role/new-role.component.spec.ts
```diff
@@ -60,22 +60,6 @@ describe('NewRoleComponent', () => {
...
-          [{ type: PreconditionType.Role, conditions: ['a', 'b', 'c', 'd'] }],
-        ],
-      }),
-    });
-    component.removePreconditionRole(index);
-    expect(
-      component.roleForm.get('data').get('enrolmentPreconditions').value[0]
-        .conditions
-    ).toEqual(['a', 'c', 'd']);
-  });
```
jrhender, line 77: Do these tests no longer make sense in Angular 14?

src/app/routes/registration/request-claim/request-claim.component.ts
```diff
@@ -276,9 +281,10 @@ export class RequestClaimComponent implements OnInit, SubjectElements {
         registrationTypes: enrolForm.registrationTypes,
       } as any);
 
+      // TODO: remove any, add proper typing to the form
       this.displayAlert(
         'Request to enrol as ' +
-          this.roleTypeForm.value.roleType.name.toUpperCase() +
+          (this.roleTypeForm.value.roleType as any).name.toUpperCase() +
```
jrhender: Would the way to fix this be to have `roleType` in the `IRoleTypeForm` interface be `{ name: string }`?

src/app/shared/expiration/date-picker/expiration-date.component.ts
```diff
@@ -91,7 +91,7 @@ export class ExpirationDateComponent implements OnInit, OnDestroy {
     }
   }
 
-  private calcSeconds(value: string): number {
+  private calcSeconds(value: Date): number {
     const d = new Date(value);
```
jrhender: Can this line be removed if the param type is `Date`?
```suggestion
  private calcSeconds(value: Date): number {
```
By the way, is this change related to the angular 14 upgrade?

jrhender (commented):
> there are few as any because Angular v14 supports typed form. However we have some forms created on the fly, which creates issues with typing. If you prefer i can add unit tests + refactor/rewrite those forms to have proper typing. Do you want from me to rewrite in this PR or should I leave it as it is?

In my opinion, we can leave it as is (with the `any`s) and add typings in another PR. My rationale is that we didn't have typings before so upgrading without typings doesn't leave us in a worse state and lets us have smaller PRs. How does that sound?

angular.json
dawidgil: No, it is not related to angular 14. I should revert this change, as I fixed this in source maps for sentry. I wrongly updated this file previously, so running `ng serve` was starting local version with `prod` flag.

src/app/routes/applications/new-role/new-role.component.spec.ts
dawidgil: Angular was complaining on those forms in tests. Also for me, those tests doesn't make sense, because they are not testing component but a specific function. Imo we shouldn't create forms like this in spec files.

src/app/shared/expiration/date-picker/expiration-date.component.ts
dawidgil: yes, it was complaining on `string` here. 
Checked: it is NOT safe to remove that Date creation. User can pass string to the input (he doesn't need to use callendar). If he pass string, then we do not have Date object, we have string. I will adjust typing that it might be `Date | string`

src/app/routes/registration/request-claim/request-claim.component.ts
dawidgil: Ok, found out that this roleType is `IRole` from icl. I won't add that to `IRoleTypeForm` as we need only `name` property. So i will just set `name` from `IRole` and there is no need to have `name` in form.

src/app/routes/applications/new-role/new-role.component.spec.ts
jrhender: Ok, sounds good

jrhender (approved):
Thanks, looks good

PR #867, opened 2023-03-20
chore: log field list for debug

whitneypurdum:


whitneypurdum:
@jrhender  @dawidgil  This is where the stringification is happening for the issuer fields. They are being stringified as expected in dev, but not in stg. The code is the same (I did a compare between branches).  Wanting to see if maybe there is a difference in role field types somehow between dev and stg - maybe something in how roles were created? If the code is identical, I'm a little stumped.

jrhender (approved):

PR #869, opened 2023-03-22
fix: build scripts

dawidgil:
PR fixes problem with `--prod` flag while building app. This is no longer supported by angular cli.


jrhender (commented):
Is the `--aot` flag also not needed? It seems it has been removed as well

dawidgil:
it is true by default: https://angular.io/guide/aot-compiler#choosing-a-compiler

jrhender (approved):

nichonien (approved):

PR #871, opened 2023-03-23
Update to angular to v15

dawidgil:
feat for task: https://energyweb.atlassian.net/browse/SWTCH-2209
App uses legacy components and legacy styling from angular material. It would be good to move to new version of angular materials in future as loose support.


jrhender (approved):
Looks good to me, thanks @dawidgil . Exciting stuff!
I wonder if we should update the `node` and `npm` versions in `package.json` after this PR?

PR #873, opened 2023-04-18
fix: json schema validation in enrolment form

dawidgil:
fix for: https://energyweb.atlassian.net/browse/SWTCH-2305
Tested with similar json schema.
Now it validates fields with empty object (it is set probably by json-editor) at setting up view.
Also it checks validation in every json schema field. 

https://www.loom.com/share/83179102aea349c4a9e3e30b78825510

### DIFF
--- a/src/app/routes/registration/enrolment-form/enrolment-form.component.ts
+++ b/src/app/routes/registration/enrolment-form/enrolment-form.component.ts
@@ -43,6 +43,8 @@ export class EnrolmentFormComponent implements EnrolmentForm {
     fields: new FormArray([]),
   });
 
+  private ajv = new Ajv();
+
   @Input() showSubmit = true;
   @Input() namespaceRegistrationRoles: Set<RegistrationTypes>;
   get fieldList(): IFieldDefinition[] {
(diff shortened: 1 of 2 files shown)


src/app/routes/registration/enrolment-form/enrolment-form.component.ts
```diff
@@ -121,18 +134,31 @@ export class EnrolmentFormComponent implements EnrolmentForm {
 
   private buildEnrolmentFormFields() {
     const values = this.enrolmentForm.value.fields;
-    return this.fieldList.map((field, index) => ({
-      key: field.label,
-      value: field.schema ? JSON.stringify(values[index]) : values[index],
-    }));
+    return this.fieldList.map((field: IFieldDefinition, index) => {
+      if (field.schema) {
+        this.checkSchema(field.schema, {}, field.label);
```
jrhender, line 139: Why the `checkSchema` here? I suppose it doesn't hurt but I'm not sure that `this.isValidSchema` is checked after this? Is rationale that an exception will be thrown if the check fails?

src/app/routes/registration/enrolment-form/enrolment-form.component.ts
```diff
@@ -66,20 +68,31 @@ export class EnrolmentFormComponent implements EnrolmentForm {
   @Output() submitForm = new EventEmitter<EnrolmentSubmission>();
 
   private fields: IFieldDefinition[];
-  isValidSchema = true;
+  isValidSchema: Record<string, boolean> = {};
```
jrhender, line 71: In `getControl` (copied below for reference), the controls are indexed by `id`. Do you that would work here (instead of indexing by `label`)? I think that might be a bit better as I think there could, in principle, be duplicate label which might break this implementation.
```ts
getControl(id: number): FormControl {
    return (this.enrolmentForm?.get('fields') as FormArray)?.controls[
      id
    ] as FormControl;
  }
```

jrhender (commented):
A couple questions, but it's looking good

src/app/routes/registration/enrolment-form/enrolment-form.component.ts
dawidgil: it might work, but `id` is generated by angular ngFor. So it is safer to not use that for checks like this.

jrhender (approved):

src/app/routes/registration/enrolment-form/enrolment-form.component.ts
dawidgil: I'm checking schema at start, because:
1. I saw that at start approve button is enabled even if the field requires array but gets object. When i change that to array and again to object, approve button is disabled. So we need to check validation while setting up.
2. without this `isValidSchema` object won't work properly as we are creating this object while creating fields.

PR #874, opened 2023-04-19
fix: problem with navigating to org creation enrolment

dawidgil:
fix for: https://energyweb.atlassian.net/browse/SWTCH-2237

### DIFF
--- a/src/app/routes/applications/applications.component.ts
+++ b/src/app/routes/applications/applications.component.ts
@@ -7,7 +7,7 @@ import {
 } from '@angular/core';
 import { ActivatedRoute, Router } from '@angular/router';
 
-import { Subject } from 'rxjs';
+import { from, Subject } from 'rxjs';
 
 import { NewOrganizationComponent } from './new-organization/new-organization.component';
 import { ListType } from '../../shared/constants/shared-constants';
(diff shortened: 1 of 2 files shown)


src/app/routes/applications/applications.component.ts
```diff
@@ -138,11 +138,14 @@ export class ApplicationsComponent implements OnInit, AfterViewInit, OnDestroy {
...
-    this.isIamEwcOwner = await this.iamService.domainsService.isOwner({
-      domain: this.envService.rootNamespace,
-    });
-    this.orgRequestButtonText = this.envService.production
+  ngOnInit() {
+    from(
+      this.iamService.domainsService.isOwner({
+        domain: this.envService.rootNamespace,
+      })
+    ).subscribe({ next: (isOwner) => (this.isIamEwcOwner = isOwner) });
```
jrhender, line 146: What this the fix? Is it that `ngOnInit` can't be async?

dawidgil: Nah, i just slightly refactored it to make tests more readable. The problem was that UI was checking `production` property instead of `isIamEwcOwner`. Build on dev env is set with production flag as true, that's why it stopped working. This now code checks if email is defined or not. If not, then we are navigating to role enrolment.

jrhender: @dawidgil 
> This now code checks if email is defined or not.

where is this check done? Is this related to the `this.orgRequestButtonText = this.envService.orgRequestEmail` line two lines below this section?

dawidgil: yes, exactly.

jrhender:
Tested locally, looks good

jrhender (approved):

PR #875, opened 2023-04-19
Fix/displaying requestor fields

dawidgil:
fix for: https://energyweb.atlassian.net/browse/SWTCH-2186
image

### DIFF
--- a/src/app/routes/enrolment/fields/fields.component.ts
+++ b/src/app/routes/enrolment/fields/fields.component.ts
@@ -16,19 +16,11 @@ import { IFieldDefinition } from '@energyweb/credential-governance/dist/src/type
 })
 export class FieldsComponent {
   @Input() title: string;
-  @Input() isIssuerRequestView = false;
+  @Input() showMergeButton = false;
   @Input() fieldsList: KeyValue<string, string | number>[];
   @Output() fieldToCopy = new EventEmitter<KeyValue<string, string | number>>();
   @Input() issuerFields: IFieldDefinition[] | undefined;
   hasValidValue(val: string | number) {
     return val && val !== 'null';
-  }
-
-  showMergeButton(field): boolean {
-    return (
-      this.isIssuerRequestView &&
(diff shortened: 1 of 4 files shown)


jrhender:
@dawidgil I think the key requirement here is that the encrypted btc address field was too long and going over the screen. Do you know if that is fixed? I'm checking because you're screenshot in the description (thanks for adding by the way) doesn't show that field.

dawidgil:
@jrhender there is a horizontal scroll bar for each json schema. So yes, this is fixing this issue.

dawidgil:
@jrhender video here: https://www.loom.com/share/726278f7b64540cdb60485afb66170c9 however i will add small padding between displayed element and button.

src/app/routes/enrolment/fields/fields.component.ts
```diff
@@ -16,19 +17,30 @@ import { IFieldDefinition } from '@energyweb/credential-governance/dist/src/type
...
+    }
+    try {
+      JSON.parse(value);
+    } catch {
+      return false;
+    }
+    return true;
+  }
+
+  parseToJson(value: string | number): JsonObject {
```
jrhender, line 40: I think this is the correct return type because the return will be undefined if `typeof value === 'number'`
```suggestion
  parseToJson(value: string | number): JsonObject | undefined {
```

jrhender (approved):
A small suggestion but looks good

src/app/routes/enrolment/fields/fields.component.ts
dawidgil: Yeah, i added this, but it not possible to happen, because `isJsonString` won't let it in there. We have first check isJsonString and later i'm parsing it. While parsing i'm 100% sure it is a string, but typings complain, that's why i added this check.

jrhender: Ok, no problem

PR #885, opened 2023-06-09
chore(deploy): remove stedin deploy

jrhender:
Removing stedin deploy as the environment is not currently in use.

`deploy.yml` had a "Sentry Release Staging" step. This is step was executed on merge to develop branch but seemed to push using staging variable. Due to this confusion and under the assumption that it was only needed for Stedin, it has been removed as well.


nichonien (approved):

JGiter (approved):

dawidgil:
that sentry was added afaik just for stedin, probably naming (title) was wrong. I suppose it might be removed entirely, because we are getting errors/warnings from sentry on every env. It is configured in  [line 90 of cf-deploy file ](https://github.com/energywebfoundation/switchboard-dapp/pull/885/files#diff-df7a9bc19b54550894ba83dcba16eda51f76f80c497e739880c5c7cfe849ac0fR90)

jrhender:
I agree @dawidgil , it is intuitive to me that the `deploy.yaml` was just for stedin and so could be removed. I will change this PR to not add the Sentry Release code to `deploy.yaml`.

However, I only see the code that would create a sentry release in `deploy.yaml` ([this line](https://github.com/energywebfoundation/switchboard-dapp/pull/885/files#diff-df7a9bc19b54550894ba83dcba16eda51f76f80c497e739880c5c7cfe849ac0fR121)). That said, that sentry release code is unexpected to me as it is being run if workflow is for `develop` branch but is using 'environment: SENTRY_STAGING_ENV` 🤔

dawidgil (approved):

PR #887, opened 2023-06-13
Update icl@8.0.0 alpha.8

JGiter:


package.json
```diff
@@ -116,6 +116,7 @@
     "karma-jasmine-html-reporter": "~2.0.0",
     "karma-read-json": "1.1.0",
     "karma-typescript": "5.5.3",
+    "moment": "^2.29.4",
```
JGiter: `moment` was extraneously imported from `@energyweb/onchain-claims`, but now onchain-claims does not install `moment`
Also `moment` now in maintanence mode. Need to replace it

dawidgil: is it possible to remove this dependency from onchain-claims and do not add it here? moment js is a pretty big lib and i removed it because of that (tree shaking doesn't work well in it). Anyway i think moment should be in dependencies not in dev dependencies.

JGiter: onchain-claims does not need it. I added it because it is used in switchboard https://github.com/energywebfoundation/switchboard-dapp/blob/63f96fd4c1a42c47750a815cb1456b2dc9d81896/angular.json#L70
I thought that it is dev dependency, because it is used only to build

dawidgil: does the build fails when there is no momentjs? 
Anyway i think you can remove this line from angular.json, because it is not used in the code afaik.

JGiter: without moment app don't starts. I will try to remove it

JGiter: without moment I have `You are not connected to EnergyWeb Volta Chain.`

dawidgil: But this is not connected at all. This info is based on data from metamask, not moment js. 
I tested it locally on your branch by removing script and moment js and everything works fine.
Are you sure you're connected to proper chain?

JGiter: yes, I checked that. let me try again

JGiter: it works without moment. problem was with cookies

dawidgil (approved):

PR #893, opened 2023-06-26
Update prod

JGiter:


jrhender (changes requested):
Sorry @JGiter , I know it is confusing, but we shouldn't update `master` for Switchboard either. We should go directly from `develop` to `release`.

JGiter:
> Sorry @JGiter , I know it is confusing, but we shouldn't update `master` for Switchboard either. We should go directly from `develop` to `release`.

Ah, ok

dawidgil (approved):

jrhender (approved):

